PR #2, opened 2020-04-17
FHS-like install tree

TheAssassin:
This PR implements a classic FHS-like installation layout, putting binaries into `bin` and data into `share`. It is implemented in the CMake build system, so that `make install` for instance copies all data into the necessary spots.

For now, this is mostly useful for Linux systems, as the in-code support for looking up the data is platform specific.


TheAssassin:
Also fixes the submodules URLs and changes some metadata and also filenames to our new "branding".

PR #3, opened 2020-04-17
Warncleanup

bmorel:
removed unused code, and replaced unspec behaviors with what was probably intended

### DIFF
--- a/src/game/client.cpp
+++ b/src/game/client.cpp
@@ -1956,7 +1956,7 @@ namespace client
                     if(flags&(1<<k))
                     {
                         n |= p.get()<<16;
-                        if(n&0x800000) n |= -1<<24;
+                        if(n&0x800000) n |= 0xFF000000;
                     }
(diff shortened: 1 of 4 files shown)


src/game/client.cpp
```diff
@@ -1956,7 +1956,7 @@ namespace client
                     if(flags&(1<<k))
                     {
                         n |= p.get()<<16;
-                        if(n&0x800000) n |= -1<<24;
+                        if(n&0x800000) n |= 0xFF000000;
```
TheAssassin, line 1959: Does seem like that value was the plan of this undefined behavior hackery. clang-tidy is not happy about the original code. If I remember that lecture right, considering an implicit conversion to unsigned long (32-bit), we'd get `0xFF000000` as you wrote there.

Using `printf` to print the value shows the compiler would also convert those numbers this way, ending up with `0xFF000000`, so I guess this is alright.

PR #8, opened 2020-04-30
Remove global stats system

TheAssassin:
This PR removes the very tightly integrated SQLite3-based global statistics system, which has been out of use for years even on the official instance due to privacy and data protection issues that were not easily fixable.

The system was unfortunately not built as a module which then was just called through an interface, but integrated into a variety of code files. It is not guaranteed that all remains have been removed. However, the (master) protocol compatibility has been retained by filling in the gaps or simply ignoring incoming messages related to the stats.


TheAssassin:
Approved by @bmorel.

PR #9, opened 2020-05-03
Improve Menus

MoonPadUSer:
Changed the UI-Layout so it's easier to understand for new players and remove all the bloat from the general menus by moving it to advanced menus


TheAssassin:
Mostly works. There's some bugs in the latest version:

- during gameplay, there's a flashing button in the main menu that brings you to the voting menu
  - should only flash during intermission
  - there's apparently some embedded cubescript that isn't working as intended (`getvote`?)
- the server browser button should have a consistent title and look (either "find servers" or "play online", the latter of which is probably easier to understand)
  - colors are already similar
  - the size isn't quite the same, I think it wouldn't matter if the in-game one also had a larger font and icon
- too many buttons on the bottom of the server browser (this zeroconf "play local" stuff is hardly ever used, I guess it could go into "advanced")
- that "mappath" issue (probably a missing command alias)

Have you added the "compatibility layer" in Cubescript already (aka aliases mapping the old command names on the new ones)?

MoonPadUSer:
1. flashing button has been replaced by a yellow one
2. the server browser button is now consistent
3. fewer buttons at the bottom of the server list
4. mappath issue should be resolved
5. compatibility layer has been added

TheAssassin:
Looks good to me! Are there any variables which we have to set to different default values?

MoonPadUSer:
> Looks good to me! Are there any variables which we have to set to different default values, e.g., to simply the scoreboard a bit for default users?

You should set the variables scorepj and scorecn to false

TheAssassin:
Let's merge this and fix bugs later, if needed.

PR #11, opened 2020-05-10
Treat unknown OS like Linux

voidanix:
High chance it's an UNIX-like OS

Co-authored-by: Morel Bérenger

### DIFF
--- a/src/cmake/build_client.cmake
+++ b/src/cmake/build_client.cmake
@@ -12,11 +12,7 @@ if(BUILD_CLIENT)
     set(client_deps ZLIB::ZLIB SDL2::Main SDL2::Image SDL2::Mixer OpenGL::GL enet)
 
     # platform specific code
-    if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
-        find_package(PkgConfig REQUIRED)
-        pkg_check_modules(x11 x11 REQUIRED IMPORTED_TARGET)
-        list(APPEND client_deps PkgConfig::x11 rt)
-    elseif(APPLE)
+    if(APPLE)
         # build OS X specific Objective-C code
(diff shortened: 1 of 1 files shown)


src/cmake/build_client.cmake
```diff
@@ -25,6 +25,10 @@ if(BUILD_CLIENT)
         if(OSXCROSS_HOST)
             message(STATUS "osxcross build, building main.m for client")
             list(APPEND mac_client_sources xcode/main.m)
+        else()
```
TheAssassin, line 28: Why _exactly_ is this in the `APPLE` block? Does CMake consider every BSD to be macOS?

TheAssassin: You might also be able to check for *BSD by comparing `CMAKE_HOST_SYSTEM_NAME` against the string `BSD`.

voidanix: What about Solaris/Illumos or Haiku, though

TheAssassin: My main concern is why it is supposed to work if it's in an `elseif(APPLE)` block.

src/cmake/build_client.cmake
```diff
@@ -26,8 +26,11 @@ if(BUILD_CLIENT)
             message(STATUS "osxcross build, building main.m for client")
             list(APPEND mac_client_sources xcode/main.m)
         endif()
-
         list(APPEND client_sources ${mac_client_sources})
+    elseif(NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
```
TheAssassin, line 30: Can you please remove the first `if`? We have it twice in here now, but once should be enough with this `elseif` clause.

TheAssassin:
Thanks a lot!

PR #12, opened 2020-05-12
Backport enet from 2.0

voidanix:


TheAssassin:
Looks good, I'm only slightly concerned about compatibility with the servers. I'm not an enet expert, but I figure it won't hurt if I'll quickly try locally.

PR #13, opened 2020-05-13
removal of new/delete local implementations + others

bmorel:
others:

* memleak fix (reported by clang --analyze)
* replacement of swap with std::swap (in a hackish way)
* some regexes I use on my vector branch to ease refactoring


TheAssassin:
Somewhat ugly indeed. What about that `#ifdef max` stuff? We might remove it later on as well...

TheAssassin:
Thank you!

PR #14, opened 2020-05-16
Improve Menus Even More

MoonPadUSer:
Bugfix and removing bloat from Serverlist and changed the spacing between "disconnect" and "quit game" buttons


MoonPadUSer:
PLEASE DO NOT MERGE AT THIS TIME! STILL CHANGES INCOMING

MoonPadUSer:
Now someone can merge

TheAssassin:
`movieh` and `moview` are gone now, there's errors, please fix.

_______________

I don't like the new server browser very much. This "show advanced" thing is just weird there, please don't belittle users (I don't insist on removing it or so, just sharing my opinion).

The map name should be shown _always_ (no, screenshots are not enough, think of Blizzard etc.), the IP/port should also just be shown (maybe append to `serverdesc` in the standard grey, so it's more "in the background" in comparison to the colorful rest, if you know what I mean).

There should _not_ be options for either of them to hide them. Try a vanilla config directory, you won't see a preview for half the servers, e.g., tuto, newmap (on edit servers), ~80% on Blizzard, ...
The "newbies don't care" argument doesn't really apply here. Many players who come back for a second or third time already are aware of different maps, and they have to vote for them, too, also on a name base. Map names are extremely important.

Talking about the options, they should be persistent. There is nothing more annoying than having to click a checkbox on every launch. If I want to see it once, I can deactivate it. But more likely, I want to see the stuff persistently. If not, deactivating is still possible easily.

IMO the amount of servers shown is not really a useful setting, given the fact we have many servers but few players. Scrolling is just fine. Cranking up the setting to the max, the thing becomes very hard to reset, especially on smaller resolutions. Also, removing the slider helps reduce the height of the "advanced options" section, which kind of bloats the UI once activated anyway.

It's a bit of a shame that "Update interval" has never had a unit with it. As a newbie, I could assume any unit, from millis to seconds to minutes to hours. I think there is no point in having e.g., 8 seconds. How about adding like, 3-4 options, e.g., 5 seconds, 15 seconds, 1 minute, deactivated? That'll be easier to read and understand.

src/engine/serverbrowser.cpp
```diff
@@ -253,7 +253,10 @@ ICOMMAND(0, addserver, "siissss", (char *n, int *p, int *r, char *d, char *h, ch
 
 VAR(IDF_PERSIST, searchlan, 0, 0, 1);
 VAR(IDF_PERSIST, maxservpings, 0, 10, 1000);
-VAR(IDF_PERSIST, serverupdateinterval, 0, 10, VAR_MAX);
+VAR(IDF_PERSIST, serverupdateinterval, 0, 15, VAR_MAX);
+VAR(IDF_PERSIST, serverbrowser_show_server_platform_and_branch, 0, 0, 1);
```
qreeves, line 257: If this doesn't have a use inside the code, then you should probably be using a persistent alias in `config/menus/servers.cfg` like such:

```
serverbrowser_show_server_platform_and_branch = 0
setpersist serverbrowser_show_server_platform_and_branch 1
```

Either that, or someone should port the `defvar` system from RE2/Tesseract, which lets you define proper variables from scripts.

MoonPadUSer: ah, didn't know that works, I'll try it out and thank you very much :D !!

TheAssassin:
Another very useful set of changes! Thanks a lot @MoonPadUSer!

PR #16, opened 2020-05-18
Add hardening by default and extra hardening option

voidanix:
This PR adds distro-grade hardening and an extra hardening option for the paranoid

Opened as a draft for compatibility with older compilers


src/CMakeLists.txt
```diff
@@ -3,28 +3,46 @@
 cmake_minimum_required(VERSION 3.8)
 
 # set compile options
-set(COMMON_C_FLAGS "-Wall -fsigned-char")
-set(COMMON_CXX_FLAGS "-fno-rtti -ffast-math")
+set(COMMON_CXX_FLAGS "-fno-rtti -ffast-math -fexceptions -Wall -fsigned-char")
```
TheAssassin, line 6: Why add `-fexceptions`? It's pretty much the default, and I just recently removed the `-fno-exceptions` since we are moving towards using more STL.

You are aware of the fact that we use plain C code, too, like e.g., enet (and previously, sqlite3)? That's where the `C_FLAGS` were used, too (which I think I took from the previous "build system" back in the days). Now, those don't see the flags any more. Any idea what the side effects will be?

src/CMakeLists.txt
```diff
@@ -3,28 +3,46 @@
...
 
 if(CMAKE_BUILD_TYPE MATCHES RELEASE)
     message(STATUS "Release build detected, adding release flags")
-    set(COMMON_C_FLAGS "${CMAKE_CXX_FLAGS} -O3 -fomit-frame-pointer")
+    set(COMMON_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O2 -fomit-frame-pointer -fstack-protector-all -fpic -fpie -D_FORTIFY_SOURCE=2")
```
TheAssassin, line 10: `-fpic` and `-fpie` don't really add any advantage here, right? I mean, they're not needed normally anyway. Why include them?

What's the advantage of using `-O2` over `-O3` like the original project did?

I don't really know what `-D_FORTIFY_SOURCE` does.

I would suggest to add a few comments above this line, explaining the flags.

src/CMakeLists.txt
```diff
@@ -3,28 +3,46 @@
...
+    if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+        set(COMMON_CXX_FLAGS "${ADDITIONAL_BUILD_FLAGS} -D_GLIBCXX_ASSERTIONS")
```
TheAssassin, line 12: That only really helps when building against libstdc++, I doubt libc++ will pick it up.

src/CMakeLists.txt
```diff
@@ -3,28 +3,46 @@
...
+    # TODO: "-Wl,-pie"
+    set(CMAKE_EXE_LINKER_FLAGS "-Wl,-z,relro -Wl,-z,now -Wl,-z,defs -pie")
 endif()
 
 option(PEDANTIC "Enable pedantic mode")
 if(PEDANTIC)
     message(STATUS "Enabling pedantic mode")
     if(CMAKE_CXX_COMPILER_ID MATCHES GNU)
-        set(COMMON_C_FLAGS "${ADDITIONAL_BUILD_FLAGS} -Wextra")
+        set(COMMON_CXX_FLAGS "${ADDITIONAL_BUILD_FLAGS} -Wextra")
```
TheAssassin, line 22: See above, the idea was to have these extra pedantic flags also for the C sources.

src/CMakeLists.txt
voidanix:
```
-fexceptions is recommended for hardening of multi-threaded C and C++ code. Without it, the implementation of thread cancellation handlers (introduced by pthread_cleanup_push) uses a completely unprotected function pointer on the stack. This function pointer can simplify the exploitation of stack-based buffer overflows even if the thread in question is never canceled.
```
The quote above comes from somebody smarter than me at Red Hat

As for the `C_FLAGS`, I didn't realize that enet still relies on them, so was a mistake on my side

src/CMakeLists.txt
voidanix: ASLR/ASR doesn't work without PIE/PIC, it's an essential mitigation

`-O2` was picked over `-O3` for `FORTIFY_SOURCE` as it relies on the optimization and `O3` is too risky for unexpected behaviour with that flag, `O2` should also help on architectures like ARM because of the reduced instruction cache footprint

`FORTIFY_SOURCE` is used for runtime buffer overflow detection

src/CMakeLists.txt
TheAssassin: Well, I'm not saying it necessarily needs any of these flags, but it might make sense to keep things consistent. This is anyway not good style, as we hack a global CMake variable. We really should do this in the `add_redeclipse_executable` macro...

src/CMakeLists.txt
TheAssassin: I don't know whether either really has performance improvements over the other. We should maybe test that. Arguably, `-O2` is pretty good already.

src/CMakeLists.txt
```diff
@@ -3,28 +3,46 @@
...
 
 if(CMAKE_BUILD_TYPE MATCHES RELEASE)
     message(STATUS "Release build detected, adding release flags")
-    set(COMMON_C_FLAGS "${CMAKE_CXX_FLAGS} -O3 -fomit-frame-pointer")
+    set(COMMON_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O2 -fomit-frame-pointer -fstack-protector-all -fpic -fpie -D_FORTIFY_SOURCE=2")
+    if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+        set(COMMON_CXX_FLAGS "${ADDITIONAL_BUILD_FLAGS} -D_GLIBCXX_ASSERTIONS")
+    endif()
+    # TODO: "-Wl,-pie"
+    set(CMAKE_EXE_LINKER_FLAGS "-Wl,-z,relro -Wl,-z,now -Wl,-z,defs -pie")
```
TheAssassin, line 15: This is also a pretty cumbersome list of linker flags. Please consider adding a docstring.

src/CMakeLists.txt
```diff
@@ -3,28 +3,46 @@
...
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${COMMON_C_FLAGS}")
+option(HARDENED "Enable hardening options")
+if(HARDENED)
+    message(STATUS "Enabling hardening options")
+    if(CMAKE_CXX_COMPILER_ID MATCHES GNU)
+	set(COMMON_CXX_FLAGS "${ADDITIONAL_BUILD_FLAGS} -fstack-clash-protection")
+    elseif(CMAKE_CXX_COMPILER_ID MATCHES Clang AND CMAKE_CXX_COMPILER_VERSION GREATER 10)
+        set(COMMON_CXX_FLAGS "${ADDITIONAL_BUILD_FLAGS} -fstack-clash-protection -fsanitize=safe-stack -mretpoline -mspeculative-load-hardening -mlvi-cfi")
+	set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=safe-stack -mspeculative-load-hardening -mlvi-cfi")
+    elseif(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 11)
```
TheAssassin, line 38: That one only really works for clang, too, but the check is arguably incomplete. Consider adding one `if` checking for clang, then do the version check stuff within there.

src/CMakeLists.txt
```diff
@@ -3,28 +3,46 @@
...
     else()
         message(FATAL_ERROR "Pedantic mode requested, but no configuration available for compiler identification '${CMAKE_CXX_COMPILER_ID}'")
     endif()
 endif()
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${COMMON_C_FLAGS} ${COMMON_CXX_FLAGS}")
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${COMMON_C_FLAGS}")
+option(HARDENED "Enable hardening options")
+if(HARDENED)
+    message(STATUS "Enabling hardening options")
```
TheAssassin, line 32: Broken indendation. Please stick to 4 spaces in the entire block.

src/enet/CMakeLists.txt
```diff
@@ -19,6 +19,9 @@ set(CMAKE_EXTRA_INCLUDE_FILES "sys/types.h" "sys/socket.h")
 check_type_size("socklen_t" HAS_SOCKLEN_T BUILTIN_TYPES_ONLY)
 unset(CMAKE_EXTRA_INCLUDE_FILES)
 
+# C files still around, fetch flags from main configuration
+set(CMAKE_C_FLAGS "${CMAKE_CXX_FLAGS}")
```
TheAssassin, line 23: Now that isn't really the solution I had in mind. Doesn't the C compiler then complain again about C++ flags?

Also, this overwrites all user-set C flags.

Please restore the original snippet, which took into account externally specified flags and the differentiation between C and C++ flags (if the latter is not needed any more, maybe move to just using `COMMON_FLAGS` and then apply them properly).

PR #22, opened 2020-05-24
Windows/macOS Release builds for CI, respect nproc on macOS

voidanix:


TheAssassin:
There is one problem with the first half of this PR. On Travis CI, you can have 5 concurrent builds. Right now, we have 4. This PR adds another 2, effectively doubling the build time (best case).

I'm not sure if release builds on the other two are therefore a good idea. Right now, we still have rather small build times, which are also a great feature.

voidanix:
Maybe it's a better idea to keep only Release builds on Travis? It passes more flags and seems to generate more warnings compared to its Debug counterpart

TheAssassin:
I was also thinking about that, given the latest changes in the build system. Maybe we just turn it around and have one Ubuntu debug build and the rest as release builds, just to have one debug build on the CI at least?

PR #28, opened 2020-05-27
replace some heap allocations with stack ones

bmorel:
3 benefits: less syscalls, less risks of memory management errors, and less memgrind whinning

### DIFF
--- a/src/game/hud.cpp
+++ b/src/game/hud.cpp
@@ -1378,7 +1378,7 @@ namespace hud
     {
         int index = POINTER_NONE;
         if(hasinput()) index = !hasinput(true) || commandmillis > 0 ? POINTER_NONE : POINTER_GUI;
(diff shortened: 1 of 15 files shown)


src/game/hud.cpp
```diff
@@ -1415,7 +1415,7 @@ namespace hud
 
     bool showname()
     {
-        if(game::focus != game::player1)
+        if(game::focus == &game::player1)
```
TheAssassin, line 1418: This clearly looks wrong. Why is that suddenly an equals comparison?

bmorel: guess a hasty replace I didn't noticed in my own code re-read.

bmorel:
buggy patch, I'll reopen one. easier to fix patches from local anyway.

Issue #25, opened 2020-05-26
Review intellectual property and copyright violations


qreeves:
As per the [Red Eclipse Mark Policy](https://github.com/redeclipse/base/blob/master/doc/trademark.txt) (a copy of which you have in this repository), I hereby ask that you respect the intellectual property and rights of the Red Eclipse project and remove our copyrighted materials.

- The name "Red Eclipse Legacy" is not a permitted use of the "Red Eclipse" name, as it is likely to cause confusion with the false belief it is an official product by the Red Eclipse team and in turn harms the community. Part of our plan included re-releasing v1.6 on Steam under the name "Red Eclipse Legacy" at a later point (and is already available on GameJolt under that branding), which has now been confused by the existence of this improperly named fork.

- The files "icon.png" and "logo.png", along with all derivatives and imagery, are provided under the provisions of the Red Eclipse Mark Policy. The use of these images in this project is not only illegal, but it serves to further confuse people as to the nature of the project.

I don't much appreciate people trying to piggyback on the hard work I've put in over the last decade. A lot of time and effort has gone into establishing the Red Eclipse brand and related intellectual properties, and I've been more than generous in making sure everything you need to make a fork is free and open source - but I had one simple rule: _please don't use our name and logos_. You of all people should have respected that.

I'm trying not to get upset and do the right thing by coming and asking nicely. I'm not interested in an argument or a debate on the issue. Should the matter not be resolved within 10 business days (close of business 11th June 2020), I will unfortunately have to take further action in order to protect the interests of the Red Eclipse project.

Lyberta:
[There is no such thing as intellectual property.](https://www.gnu.org/philosophy/not-ipr.html) It is a lie perpetuated by evil and greedy people who try to equate copyright infringement (among other things) to theft which is nonsense. You only have trademark rights.

TheAssassin:
Hey @qreeves, thanks for starting a dialog first. We are in the process of renaming, but due to lack of personpower, this'll take a couple of days from now. We had set up a public poll [in our wiki](https://github.com/redeclipse-legacy/base/wiki/Name-Suggestions) and this poll closed yesterday, now internal review and discussion is in process. The name "Red Eclipse Legacy" will, along with the logo and name, be replaced another one. Your deadline is roughly 2 weeks away, we should be done by then. We can keep this issue updated.
[...]

PR #29, opened 2020-05-29
Apply new name, mk. 1

TheAssassin:
Contributes to #25.

This PR changes the name in a lot of places in this repository. It also removes additional unused or otherwise obsolete or deprecated files.

It took a few hours to do all the changes semi-automatically and review all changes and commit them properly. It's a lot for you to review, but I'd appreciate if you could do it with great care so you can spot potential issues before we merge the PR.

We still have to apply the name changes to all data submodules. Also, we need to remove the trademark policy from there. We're not Red Eclipse and make no trademark claims of any kind for our name, Blue Nebula.

The logos etc. have not been changed yet, since we first need to either find someone to draw one (ideally in SVG), or otherwise change the logo. Also, no color changes have been made so far; at a later point, we should also change the red-ish scheme to something blue-ish, which'll fit our name better.


bin/redeclipse.app/Contents/Info.plist
```diff
@@ -5,15 +5,15 @@
 	<key>CFBundleDevelopmentRegion</key>
 	<string>English</string>
 	<key>CFBundleName</key>
-	<string>Red Eclipse Legacy</string>
+	<string>Blue Nebula</string>
```
voidanix, line 8: Can you rename `bin/redeclipse.app` to `bin/blue-nebula.app`?

src/xcode/redeclipse-Info.plist
```diff
@@ -5,9 +5,9 @@
 	<key>CFBundleDevelopmentRegion</key>
 	<string>English</string>
 	<key>CFBundleExecutable</key>
-	<string>redeclipse</string>
+	<string>blue-nebula</string>
```
voidanix, line 8: `src/xcode/redeclipse.xcodeproj/project.pbxproj` still references RE; can you remove the entire `xcode` folder, along with `src/macbuild.sh` as cmake is used?

`src/blue-nebula.cbp` could also be removed/edited

src/engine/server.cpp
```diff
@@ -1590,7 +1590,7 @@ void setlocations(bool wanthome) {
         }
 
         strncpy(fhs_share_dir, bin_dir, PATH_MAX - 1);
-        strncat(fhs_share_dir, "/../share/redeclipse-legacy/", PATH_MAX - 1 - strlen(fhs_share_dir));
+        strncat(fhs_share_dir, "/../share/blue-nebula/", PATH_MAX - 1 - strlen(fhs_share_dir));
```
voidanix, line 1593: Line 1575 references RE

src/scripts/servinit-defaults
```diff
@@ -4,7 +4,7 @@
 # doc/examples/servexec.cfg
 #
 # It reads all sv_* variables in doc/examples/servexec.cfg, extracts their
-# default value by executing them in a Red Eclipse Legacy server, then feeds
+# default value by executing them in a Blue Nebula server, then feeds
```
voidanix, line 7: See line 16

bin/redeclipse.app/Contents/Info.plist
TheAssassin: I have not touched this `.app` stuff intentionally, we're not using it at the moment anyway. We need to decide first how to proceed with packaging for macOS before we should touch those files IMO.

src/xcode/redeclipse-Info.plist
TheAssassin: `macbuild.sh` is gone already. I think we also won't need these frameworks anymore, they're annoying to work with anyway. The Code::Blocks project is gone, too.

src/engine/server.cpp
TheAssassin: Will be fixed, thanks.

src/scripts/servinit-defaults
TheAssassin: The script is broken anyway, since we don't have these scripts any more. It should be fixed when ever needed again to extract the variables, I guess.

TheAssassin:
> The Multiplayer guidelines menu appears and doesn't display anything, showing just accept/decline; showing something like This text is work in progress would be good

Should be fixed by @bmorel's PR #30... I think. We'll see.

> Maybe I'm nitpicking but blue-nebula seems too long to display as the branch in-game, wouldn't nebula or just blue be better?

Fix is on the way.

PR #37, opened 2020-06-03
Add simple name auto-completion for the chat

MoonPadUSer:


MoonPadUSer:
Todo: exclude bots from the auto-completion

bmorel:
Things that would block me accept this:

* no sanity check on s: there is no certainty you don't write out of memory thus I consider this code unsafe and not acceptable.

Now, general coding sanity:

* Instead of using if/else that takes the whole function and change indentation, just return at end of if.
* Indentation change found in 2nd patch near like 824, easy to see with a diff.
* lines ending with spaces found
* what is the printf( "start" ...) meant to do in a part of the code you don't touch?
* strlen( foo ) in a loop that never changes foo is slow
* strlen is slow anyway, avoid calling it for nothing (I see at least 1 useless call)

More specific to that current code:

* avoid using the loop macros, use for( auto const& foo : foobar ) or normal macros.
* _please_ stop having for loops and if condition on same line to then use misleading indentation.
* complete() is used only once in the whole code. Take the occasion to have the architecture safer, for example by using an std::string (there are other methods, too)
* for the same reason, there's no point in keeping it's declaration extern or in a header anyway
* avoid using an external structure's variables and literal constants. Instead, add it a method with a name that describes what it does (for example here: 'bool clientstate::is_player(void) const;' and 'char const* clientstate::name(void) const;'

MoonPadUSer:
what the actual fuck? Travis CI is done but it still shows that it's not done here??

src/engine/console.cpp
```diff
@@ -772,15 +775,57 @@ void addlistcomplete(char *command, char *list)
 COMMANDN(0, complete, addfilecomplete, "sss");
 COMMANDN(0, listcomplete, addlistcomplete, "ss");
 
-void complete(char *s, const char *cmdprefix)
+void complete(char *s, size_t s_size, const char *cmdprefix)
 {
+
+    char* name = strrchr(s, '@');
+
+    if (name) // autocomplete name
```
TheAssassin, line 783: Inline comments. Please avoid.

PR #40, opened 2020-06-07
Add option to hide incompatible servers from serverbrowser

MoonPadUSer:
This PR adds a new checkbox to the Serverbrowser options called "Hide Incompatible Servers" which will ... well, hide the incompatible servers in the serverbrowser. I know, it ain't much but I think it's nice to have


TheAssassin:
Thank you!

PR #51, opened 2020-06-08
Customizable compass menu

### LINKED ISSUE #45: Custom entries for circle menu
I'd like to give players the ability to modify a custom circle menu with the entries they want

MoonPadUSer:
Adds new tab to the User Interface Options:
NewTab
You can modify up to 6 shortcuts by setting a name and entering any cubescript command in the action field like this:
ShowcaseCommand
In-Game you can then press the C key (rebindable) and it'll pop up a compass menu:
OpenMenu
they work exactly like the voiceline compass menus, so when you press a shortcut (in this case: "Hello World!") and it'll execute the command you entered before
Demonstration

Closes #45.

### DIFF
--- a/src/engine/menus.cpp
+++ b/src/engine/menus.cpp
@@ -11,6 +11,19 @@ struct menu;
 static guient *cgui = NULL;
 static menu *cmenu = NULL;
 
+SVAR(IDF_PERSIST, customcompassentryname1, "");
+SVAR(IDF_PERSIST, customcompassentryname2, "");
+SVAR(IDF_PERSIST, customcompassentryname3, "");
+SVAR(IDF_PERSIST, customcompassentryname4, "");
+SVAR(IDF_PERSIST, customcompassentryname5, "");
+SVAR(IDF_PERSIST, customcompassentryname6, "");
+SVAR(IDF_PERSIST, customcompassentryaction1, "");
+SVAR(IDF_PERSIST, customcompassentryaction2, "");
+SVAR(IDF_PERSIST, customcompassentryaction3, "");
+SVAR(IDF_PERSIST, customcompassentryaction4, "");
+SVAR(IDF_PERSIST, customcompassentryaction5, "");
+SVAR(IDF_PERSIST, customcompassentryaction6, "");
+
 struct menu : guicb
 {
     char *name, *header;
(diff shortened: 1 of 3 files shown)


TheAssassin:
Not sure if I like this very much. I'd make it strictly opt-in. Most people, including me, probably won't want to use this.

Also, I'm not so sure about the default bind `c`. I know tons of players using `c` to crouch, they're used to this from many other games. We need to find an alternative.

MoonPadUSer:
> Not sure if I like this very much. I'd make it strictly opt-in. Most people, including me, probably won't want to use this.
> 
> Also, I'm not so sure about the default bind `c`. I know tons of players using `c` to crouch, they're used to this from many other games. We need to find an alternative.

1. I am not sure if I get what you mean. It is as opt-in as the normal voiceline compass menus, you don't have to use it if you don't want to, it's just about giving people the opportunity to bind anything to a button in this compass menu
2. I agree with you that `C` is not the best keybind. I'll do some testing and see what other keybinds are available.

src/engine/menus.cpp
```diff
@@ -11,6 +11,19 @@ struct menu;
 static guient *cgui = NULL;
 static menu *cmenu = NULL;
 
+SVAR(IDF_PERSIST, customcompassentryname1, "");
```
bmorel, line 14: new vars. No doc.

src/engine/menus.cpp
```diff
@@ -737,6 +755,7 @@ COMMAND(0, ui_tooltip, "si");
 COMMAND(0, ui_textfield, "sisbisisi");
 COMMAND(0, ui_keyfield, "sisbissi");
 COMMAND(0, ui_editor, "siiibisss");
+COMMAND(0, ui_execute , "s");
```
bmorel, line 758: new command, no doc. Why not using execute() directly?

MoonPadUSer: I think I can actually remove this entire thing

src/engine/menus.cpp
MoonPadUSer: I think I can remove those too.... xD

MoonPadUSer:
I am going to close this PR because it's outdated and I'd rather rewrite it entirely since there are many things that can be improved a lot

PR #57, opened 2020-06-09
Add Ability to display system time in the HUD and menus

MoonPadUSer:
Adds this toggle:
Radiobutton
that enables showing the current time in the lower right corner of the screen:
Main Screen:
Homescreen
In-Game:
In-Game
In the code I add drawing the system time in the `drawhud(...)` and the `drawbackground(...)` functions so it'll be drawn inside the menus (`drawbackground`) and in-game (`drawhud`).
One might argue that the time is pretty small, and I agree, but that is a good thing in my opinion, that way it doesn't distract gameplay but you'll still be able to keep an eye on the time :D

### DIFF
--- a/config/menus/options.cfg
+++ b/config/menus/options.cfg
@@ -668,13 +668,13 @@ options_radar = [
         ui_font "little" [ ui_text "^fa(default: compass-distance)" ]
     ]
     ui_list [
-        ui_radiobutton2 "compass" radarstyle 0
+        ui_radiobutton2 "Compass" radarstyle 0
         ui_strut 1
-        ui_radiobutton2 "compass-distance" radarstyle 1
+        ui_radiobutton2 "Compass-Distance" radarstyle 1
         ui_strut 1
-        ui_radiobutton2 "full-hud" radarstyle 2
+        ui_radiobutton2 "Full HUD" radarstyle 2
         ui_strut 1
-        ui_radiobutton2 "corner" radarstyle 3
+        ui_radiobutton2 "Minimap" radarstyle 3
     ]
     ui_strut 0.5
     ui_text "filter radar by:"
@@ -733,13 +733,13 @@ options_radar = [
         ] "show player names" [
             ui_text "toggles visibility of player names" "" 0x22FF22 "" 1200
         ]
-        cases $guirollovername "compass" [
+        cases $guirollovername "Compass" [
             ui_text "compass styled radar showing the direction of targets (classic style)" "" 0x22FF22 "" 1200
-        ] "compass-distance" [
+        ] "Compass-Distance" [
             ui_text "compass styled radar showing the direction and relative distance of targets" "" 0x22FF22 "" 1200
-        ] "full-hud" [
+        ] "Full HUD" [
             ui_text "compass styled radar showing the direction and relative distance of targets with overlaid points of interest" "" 0x22FF22 "" 1200
-        ] "corner" [
+        ] "Minimap" [
             ui_text "top-down view styled radar in the top-right corner of the screen showing actual positions of targets" "" 0x22FF22 "" 1200
         ]
         cases $guirollovername "off" [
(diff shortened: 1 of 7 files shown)


config/menus/options.cfg
```diff
@@ -421,6 +421,12 @@ options_statistics = [
         ui_strut 1
         ui_radiobutton2 "more verbose" showfps 3
     ]
+    ui_text "On-Screen System Time:"
```
TheAssassin, line 424: Include *where* it will pop up, please.

src/engine/main.cpp
```diff
@@ -911,6 +912,7 @@ int main(int argc, char **argv)
     #endif
     #endif
 
+    setlocale( LC_ALL, "" );
```
TheAssassin, line 915: I am absolutely no fan of such code. It is very likely to interfere with our localization plans.

Why is this needed? It's not even documented...

src/game/game.h
```diff
@@ -1683,6 +1683,8 @@ namespace hud
     extern int raceinventory(int x, int y, int s, float blend);
     extern int drawscore(int x, int y, int s, int m, float blend, int count);
     extern void cleanup();
+    extern char const* system_time;
+    extern bool show_system_time();
```
TheAssassin, line 1687: You should be able to get away without some global static variable. This one is particularly annoying because it's polluting _all_ files' namespace.

src/game/hud.cpp
```diff
@@ -7,6 +7,7 @@ namespace hud
 {
     const int NUMSTATS = 11;
     int damageresidue = 0, hudwidth = 0, hudheight = 0, lastteam = 0, laststats = 0, prevstats[NUMSTATS] = {0}, curstats[NUMSTATS] = {0};
+    char const* system_time = nullptr;
```
TheAssassin, line 10: See above. Should go away. _At least_ from the global namespace.

src/game/hud.cpp
```diff
@@ -39,6 +40,7 @@ namespace hud
     VAR(IDF_PERSIST, showloadingversion, 0, 1, 1);
     VAR(IDF_PERSIST, showloadingurl, 0, 1, 1);
 
+    VAR(IDF_PERSIST, showsystemtime, 0, 0, 1);
```
TheAssassin, line 43: Aren't HUD options prefixed `hud` or so?

src/game/hud.cpp
```diff
@@ -698,6 +700,11 @@ namespace hud
         result(list.getbuf());
     });
 
+    bool show_system_time()
```
TheAssassin, line 703: This getter is kind of pointless given the fact the variable is accessible globally already.

src/game/hud.cpp
```diff
@@ -3156,6 +3163,12 @@ namespace hud
                     default: break;
                 }
             }
+            if (showsystemtime)
```
TheAssassin, line 3166: Blank lines in between these blocks wouldn't hurt anyone.

src/game/hud.cpp
```diff
@@ -3156,6 +3163,12 @@ namespace hud
...
+            {
+                pushfont("console");
+                cy[1] -= draw_textf("%s", cx[1], cy[1], 0, 0, 255, 255, 255, bf, TEXT_RIGHT_UP, -1, bs, 1, system_time);
+                popfont();
+            }
```
TheAssassin, line 3171: Another blank line below this makes the reviewer happy.

config/menus/options.cfg
MoonPadUSer: That would be inconsistent looking at the rest of the interface, e.g. On-Screen FPS

src/engine/main.cpp
MoonPadUSer: to be completely honest, I am not sure if it is NEEDED, I just thought it couldn't hurt anybody. I'll remove it and see if it still works

MoonPadUSer:
I'll remove the variables completely and make one function that gives back the correctly formatted time

src/game/hud.cpp
MoonPadUSer: nope, they're not

PR #58, opened 2020-06-09
Add ability to use Alt + Backspace to delete entire words in the console

### LINKED ISSUE #54: Alt + Backspace in the Console to delete entire words
The title should describe it pretty well I think. #58

MoonPadUSer:
Hold LALT and press backspace to delete an entire word in the console

Closes #54.

### DIFF
--- a/src/engine/console.cpp
+++ b/src/engine/console.cpp
@@ -512,16 +512,20 @@ bool consolekey(int code, bool isdown)
                 if (holding_alt)
                 {
                     char* last_word = strrchr(commandbuf, ' ');
-                    if (last_word)
+                    if (last_word != NULL)
                     {
-                        commandbuf[(commandbuf_len - strlen(last_word))] = 0;
+                        // add 1 to the len to only remove the word without the space, as this will be removed later
+                        size_t new_len = (last_word - commandbuf) + 1;
+                        commandbuf[new_len] = '\0';
+                        commandbuf_len = new_len;
                     }
                     else
                     {
-                        commandbuf[0] = 0;
+                        commandbuf[0] = '\0';
+                        commandbuf_len = 0;
                     }
                 }
-                
+
 
                 int len = (int)commandbuf_len, i = commandpos>=0 ? commandpos : len;
                 if(i<1) break;
--- a/src/engine/console.cpp
+++ b/src/engine/console.cpp
@@ -12,6 +12,7 @@ bigstring commandbuf;
 char *commandaction = NULL, *commandicon = NULL;
 enum { CF_COMPLETE = 1<<0, CF_EXECUTE = 1<<1, CF_MESSAGE = 1<<2 };
 int commandflags = 0, commandpos = -1, commandcolour = 0;
+bool holding_alt = false;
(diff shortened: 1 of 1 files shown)


src/engine/console.cpp
```diff
@@ -506,7 +507,21 @@ bool consolekey(int code, bool isdown)
 
             case SDLK_BACKSPACE:
             {
-                int len = (int)strlen(commandbuf), i = commandpos>=0 ? commandpos : len;
+                size_t commandbuf_len = strlen(commandbuf);
+                if (holding_alt)
```
TheAssassin, line 511: Leave a blank line above, then insert a comment above the `if` explaining why this branch is added here.

src/engine/console.cpp
```diff
@@ -506,7 +507,21 @@ bool consolekey(int code, bool isdown)
...
+                {
+                    char* last_word = strrchr(commandbuf, ' ');
```
TheAssassin, line 513: Are you sure you're handling the empty string correctly?

src/engine/console.cpp
```diff
@@ -506,7 +507,21 @@ bool consolekey(int code, bool isdown)
...
+                    {
+                        commandbuf[(commandbuf_len - strlen(last_word))] = 0;
+                    }
+                    else
+                    {
+                        commandbuf[0] = 0;
+                    }
+                }
+
+                int len = (int)commandbuf_len, i = commandpos>=0 ? commandpos : len;
```
TheAssassin, line 524: Consider adding more blank lines to group what's logically related. I'd add one below here.

src/engine/console.cpp
MoonPadUSer: Not anymore, since you now said smth. Is there a better way to do this?

src/engine/console.cpp
MoonPadUSer: done.

src/engine/console.cpp
TheAssassin: Your code is not very complex, so try it in a separate `.cpp` file in an MWE. There you can call your code on different strings, and see if it crashes.

TheAssassin:
Note for self: PR can and should be squashed.

src/engine/console.cpp
TheAssassin: Please feel free to just mark conversations as resolved when you're done.

src/engine/console.cpp
MoonPadUSer: Why should it crash tho, I can't see any reason why it would

src/engine/console.cpp
```diff
@@ -506,7 +507,23 @@ bool consolekey(int code, bool isdown)
...
+                    {
+                        commandbuf[(commandbuf_len - strlen(last_word))] = 0;
+                    }
+                    else
+                    {
+                        commandbuf[0] = 0;
+                    }
+                }
+                
+
```
TheAssassin, line 525: Now, there's two blank lines here, and none above the comment above the if...?

PR #70, opened 2020-06-17
Cleanup getservers and rename serverinfo_ variables

MoonPadUSer:
This PR is only for improving the readability of `config/menus/servers.cfg` by separating the function `getserver` into `getserver_title`, `getserver_port`, `getserver_version_branch`, etc.


MoonPadUSer:
yes, I guess, not sure if that'd be slower tho

TheAssassin:
Is it highly performance critical code in your opinion?

Let me ask a different question: is this `getservers` call used anywhere else other than in this cubescript file?

MoonPadUSer:
As I said I am not entirely sure if the code is performance critical, tho it could be since these functions are called all the time also, the getserver call isn't used anywhere else. It can be done in cubescript but that would bloat the cubescript namespace like it now does the c++ namespace

MoonPadUSer:
Ok, I now moved all the functions into cubescript, there doesn't seem to be any kind of performance issue

config/menus/glue.cfg
```diff
@@ -359,3 +359,32 @@ looplist i [
     ui_checkbox2 ui_edit_button ui_edit_button2 ui_edit_checbox ui_edit_checbox2
 ] [setcomplete $i 1]
 
+get_server_count            = [ result (getserver) ]
```
TheAssassin, line 362: Naming inconsistency. This one uses `get_server`, but the rest is `getserver`. I'd really just go for `get_server`, I guess, easier to read and understand.

MoonPadUSer: done.

Issue #64, opened 2020-06-11
Manage triggers server-side in editing


Goldenbrain1:
**First seen in**: Red eclipse 1.6

**Observation**: In editing triggers works client-size and don't impact other players

**How to reproduce**:
1. Place for example a door and a trigger (let's say an toggle-action one)
2. Ask another player to stand on the top of the door
3. Open the door and see that the other player isn't impacted by the door opening

![Capture d’écran de 2020-06-11 12-50-34](https://user-images.githubusercontent.com/35398459/84378917-3697cd00-abe5-11ea-9134-a81ed11e59ce.png)

Though in an other mode than editing the triggers work fine

bmorel:
Tried with one blue-nebula and one red-eclipse 1.6 client on a red-eclipse 1.6 server. The problem seems not to be related to triggers when they are triggered by a player's action, but when their state is artificially changed by someone in edit mode.
This means there is no proof that this problem does not affect other modes, since changing a state is only possible in edit mode.
Can you confirm that I reproduced what you describe?

MoonPadUSer:
Ok, I just tested it by having 2 clients running, one hosting and one just connecting, when I created a trigger and a door with the hosting client it worked perfectly, but when I created a trigger and a door with the non-hosting client it behaved like @Goldenbrain1 described.

MoonPadUSer:
Ok, my previous testing was somehow faulty. Anyways, I figured it out and will provide a fix asap

MoonPadUSer:
# How to reproduce:
### What you need:
2 clients that are playing on the same server in editing
for this demonstration I colored the clients orange and green
### What you need to do
Step 1:
Client 1 creates trigger and connects it to mapmodel/particles/sound/etc
![Demonstration-Setup-Client](https://user-images.githubusercontent.com/37220464/85000234-69484500-b152-11ea-91c0-82116c1f4937.png)
Step 2:
Client 2 triggers the trigger
# Result:
View client 1:
![Demonstration-View-Client-Door-Open](https://user-images.githubusercontent.com/37220464/85000398-a4e30f00-b152-11ea-9437-92f4fda25c3f.png)
View client 2:
![Demonstration-Host-In-Front-Of-Door](https://user-images.githubusercontent.com/37220464/85000429-add3e080-b152-11ea-80d3-98299bd085b1.png)

[...]

PR #72, opened 2020-06-17
Fix edit mode entity update bug

MoonPadUSer:
as described in issue #64 triggers wouldn't work on entities that you have never touched (literally, once you clicked on them, they were added to a list and they'd start working) ,I fixed this by calling the function that gets called when you edit an entity (`entities::editent(...)`) when the server sends an `N_EDITENT` event

### DIFF
--- a/src/engine/world.cpp
+++ b/src/engine/world.cpp
@@ -1183,5 +1183,6 @@ void mpeditent(int i, const vec &o, int type, attrvector &attr, bool local)
         e.attrs.add(0, max(entities::numattrs(e.type), min(attr.length(), MAXENTATTRS)) - e.attrs.length());
         loopk(min(attr.length(), e.attrs.length())) e.attrs[k] = attr[k];
         addentity(i);
+        entities::editent(i, false);
     }
 }
--- a/src/game/entities.cpp
+++ b/src/game/entities.cpp
@@ -1254,12 +1254,18 @@ namespace entities
         if(!m_edit(game::gamemode) && suicide) game::suicide(d, HIT_SPAWN);
     }
 
-    void editent(int i)
+    void editent(int i, bool local)
     {
         extentity &e = *ents[i];
-        fixentity(i, true);
-        if(m_edit(game::gamemode) && game::player1.state == CS_EDITING)
-            client::addmsg(N_EDITENT, "ri5iv", i, (int)(e.o.x*DMF), (int)(e.o.y*DMF), (int)(e.o.z*DMF), e.type, e.attrs.length(), e.attrs.length(), e.attrs.getbuf()); // FIXME
+        if (local)
+        {
+            fixentity(i, true);
+            if(m_edit(game::gamemode) && game::player1.state == CS_EDITING)
+            {
+                client::addmsg(N_EDITENT, "ri5iv", i, (int)(e.o.x*DMF), (int)(e.o.y*DMF), (int)(e.o.z*DMF), e.type, e.attrs.length(), e.attrs.length(), e.attrs.getbuf()); // FIXME
+            }
+        }
+
         if(e.type < MAXENTTYPES)
         {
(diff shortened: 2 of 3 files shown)


src/game/entities.cpp
```diff
@@ -1254,12 +1254,18 @@ namespace entities
         if(!m_edit(game::gamemode) && suicide) game::suicide(d, HIT_SPAWN);
     }
 
-    void editent(int i)
+    void editent(int i, bool local)
```
TheAssassin, line 1257: Please document the meaning of this new variable. It is not obvious at all.

src/shared/igame.h
```diff
@@ -4,7 +4,7 @@ namespace entities
 {
     extern int numattrs(int type);
     extern int triggertime(extentity &e);
-    extern void editent(int i);
+    extern void editent(int i, bool local = true);
```
TheAssassin, line 7: Same here. Variable should be documented.

src/engine/world.cpp
```diff
@@ -1183,5 +1183,6 @@ void mpeditent(int i, const vec &o, int type, attrvector &attr, bool local)
         e.attrs.add(0, max(entities::numattrs(e.type), min(attr.length(), MAXENTATTRS)) - e.attrs.length());
         loopk(min(attr.length(), e.attrs.length())) e.attrs[k] = attr[k];
         addentity(i);
+        entities::editent(i, false);
```
TheAssassin, line 1186: You need to document this call in-code. Write a comment which explains why you call this function here.

MoonPadUSer:
I changed the variable names so they are more descriptive (at least I hope they are) and added one comment. I hope this is enough but I am very unsure about it.

src/game/entities.cpp
```diff
@@ -1254,16 +1254,21 @@ namespace entities
...
-            {
-                client::addmsg(N_EDITENT, "ri5iv", i, (int)(e.o.x*DMF), (int)(e.o.y*DMF), (int)(e.o.z*DMF), e.type, e.attrs.length(), e.attrs.length(), e.attrs.getbuf()); // FIXME
-            }
+        }
+
+        if (send_update_to_server &&
+            m_edit(game::gamemode) &&
+            game::player1.state == CS_EDITING)
+        {
+            client::addmsg(N_EDITENT, "ri5iv", i, (int)(e.o.x*DMF), (int)(e.o.y*DMF), (int)(e.o.z*DMF), e.type, e.attrs.length(), e.attrs.length(), e.attrs.getbuf()); // FIXME
```
bmorel, line 1271: since #59, size() can be used for vector. Since that line is touched, why not use size() instead of length()?
Also, we're using C++, not C, take the occasion to remove the C-style casts when possible.

src/game/entities.cpp
```diff
@@ -1254,12 +1254,18 @@ namespace entities
...
         extentity &e = *ents[i];
-        fixentity(i, true);
-        if(m_edit(game::gamemode) && game::player1.state == CS_EDITING)
-            client::addmsg(N_EDITENT, "ri5iv", i, (int)(e.o.x*DMF), (int)(e.o.y*DMF), (int)(e.o.z*DMF), e.type, e.attrs.length(), e.attrs.length(), e.attrs.getbuf()); // FIXME
+        if (local)
+        {
+            fixentity(i, true);
+            if(m_edit(game::gamemode) && game::player1.state == CS_EDITING)
+            {
+                client::addmsg(N_EDITENT, "ri5iv", i, (int)(e.o.x*DMF), (int)(e.o.y*DMF), (int)(e.o.z*DMF), e.type, e.attrs.length(), e.attrs.length(), e.attrs.getbuf()); // FIXME
```
bmorel: C-style cast should be avoided. Since #59 we can use STL-compatible methods for vector in most cases.

PR #78, opened 2020-06-22
reintroduce changes that were accidentally reverted by #57 in commit cd1090f

bmorel:


bmorel:
Well, should have been obvious, a revert commit can't be cancelled by just sending anew original commit.

PR #79, opened 2020-06-23
Rework map selection menu

MoonPadUSer:
New map selection:
map_selection
New voting menu:
voting
It was originally created by IceflowRE and modified by me to be fully compatible.

Attention, this requires  in textures to be merged first


shacknetisp:
There are a couple things that might need addressed, but this looks very good. Excellent work on both your and Iceflower's parts.

- Do we need `Hide map prieviews`?
  - Also, it's misspelled.
- There should be a server-side random map selection; that is: voting for the map `<random>`

MoonPadUSer:
> There are a couple things that might need addressed, but this looks very good. Excellent work on both your and Iceflower's parts.
> 
>     * Do we need `Hide map prieviews`?
>       
>       * Also, it's misspelled.
> 
>     * There should be a server-side random map selection; that is: voting for the map `<random>`

Indeed, we need hide map previews cuz afaik cubescript isn't optimized in that way and if someone has 100+ maps it could get fairly laggy, maybe it doesn't but I'm rather safe than sorry. we'll see as time goes on

I agree, I should add "<Random>"

MoonPadUSer:
I added the random map:
Screenshot from 2020-07-01 16-11-40

MoonPadUSer:
Fixed the spelling too now

config/menus/maps.cfg
```diff
@@ -1,210 +1,88 @@
...
+
+truncate_string = [
+    if (!=s (limitstring $arg1 $arg2) $arg1) [
+        result (concatword (limitstring $arg1 (- $arg2 3)) "...")
+    ] [
+        result (limitstring $arg1 $arg2)
+    ]
+]
+
+if (getalias hidemappreviews) [] [ hidemappreviews = 0 ]
```
shacknetisp, line 14: This should be replaced with simply `hidemappreviews = 0`

The `setpersist` and `setcomplete` are fine as is, since this is an actual setting.
Consider adding a setdesc if you wish this to act similar to a normal c++ setting variable.

config/menus/maps.cfg
```diff
@@ -1,210 +1,88 @@
...
-map_path = ""
-map_extra = ""
-map_image = ""
-map_octa = 0
 if (getalias mapfavs) [] [mapfavs = ""]
 setpersist mapfavs 1
 setcomplete mapfavs 1
 if (getalias demofavs) [] [demofavs = ""]
 setpersist demofavs 1
 setcomplete demofavs 1
```
shacknetisp, line 30: Again with these, you do not need the extra code to check if they exist; config.cfg will override them anyway.

config/menus/maps.cfg
```diff
@@ -213,455 +91,681 @@ mutsvar = [
...
+									ui_font "little" [
+										if (!= $mapcategory 2) [ui_button (getmapmodemuts $i) []] [ui_button "^f[0x808080]sauerbraten" []]
+									]
+									ui_strut 25 1
+								] [
+									ui_strut 0.5
+									ui_button "<Random>"
+									ui_strut 28 1
+								]
+							]
```
shacknetisp, line 146: Needs correct indention.

These are tabs, they should be four spaces per indent.

config/menus/maps.cfg
MoonPadUSer: yep, will do

config/menus/maps.cfg
MoonPadUSer: ok, I'll add a desc

PR #80, opened 2020-06-23
Rework resolution options

MoonPadUSer:
Inside the Display & Graphics Options I removed the "Resolution" tab and fused it with "Display":
display_tab
if you click on the resolution you'll get redirected to this menu:
resolution_settings


config/menus/options.cfg
```diff
@@ -1,6 +1,9 @@
+if (getalias selectedaspectratio) [] [ selectedaspectratio = 1 ]
```
shacknetisp, line 1: Replace with simply `selectedaspectratio = 1`
The alias will be overridden in config.cfg since it is `setpersist`ed.

config/menus/options.cfg
```diff
@@ -1,6 +1,9 @@
...
+setpersist selectedaspectratio 1
+setcomplete selectedapsectratio 1
```
shacknetisp, line 3: Why do you `setcomplete`? Should people be accessing this alias from the console?

MoonPadUSer: tbh, I don't even knew what it did, but I did it because the ones beneigh it did too

config/menus/options.cfg
MoonPadUSer: alright, will do

config/menus/options.cfg
shacknetisp: `setcomplete` is used to allow various forms of tab-completion; this one makes the alias itself tab-completable

MoonPadUSer: oooh, ok that makes sense, thx

PR #82, opened 2020-06-24
Fix zoom crash/freeze.

shacknetisp:
To reproduce the freeze in master, zoom with the rifle and press `1` to switch to the pistol. `firstpersonbob` must be enabled.
This was caused by dividing by the `cookzoom` variable of the pistol, which is zero. The code assumes it is non-zero.
This can also occur when a weapon that should zoom has a `cookzoom` of zero: stopping the zoom will freeze/crash.

Alternative solution: add checks everywhere `cookzoom` is used to ensure there is no div by zero.

Examples of place where this division happens (search for uses of `cookzoom` in `game.cpp`):
https://github.com/blue-nebula/base/blob/4a47f01f239d9b580e2cf2e8da853ba97ff219ba/src/game/game.cpp#L1009-L1014
https://github.com/blue-nebula/base/blob/4a47f01f239d9b580e2cf2e8da853ba97ff219ba/src/game/game.cpp#L2304-L2310

### DIFF
--- a/src/game/weapons.h
+++ b/src/game/weapons.h
@@ -248,8 +248,8 @@ WPVARM(IDF_GAMEMOD,  cooktime, 0, VAR_MAX,
     0,          0,          0,          0,          0,          0,          0,          0,          0,          3000,       0,          0,          0,
     0,          0,          0,          0,          0,          0,          2000,       0,          650,        3000,       0,          0,          0
 );
-WPVAR(IDF_GAMEMOD,  cookzoom, 0, VAR_MAX,
-    0,          0,          0,          0,          0,          0,          0,          0,          650,        0,          0,          0,          0
+WPVAR(IDF_GAMEMOD,  cookzoom, 1, VAR_MAX,
+    1,          1,          1,          1,          1,          1,          1,          1,          650,        1,          1,          1,          1
 );
 WPFVAR(IDF_GAMEMOD,  cookzoommin, 1, 150,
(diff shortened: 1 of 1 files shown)


MoonPadUSer:
actually it looks like the game always crashes in the function `game::fixfullrange(float&, float&, bool)` which is defined in `src/game/game.cpp`

shacknetisp:
I believe that crash/freeze in fixfullrange is due to an infinite loop caused by the invalid number as a result of the code I mentioned.

TheAssassin:
I'm sorry, but I still can't reproduce the behavior on `master` (pressing 2, as that gives me a pistols, 1 is claws).

Are there side effects with regards to changing the minimum value? I'd assumed `0` means "off", whereas `1` means "very short cook time".

Right now I think it's better to make uses of `cookzoom` safe by checking for a value of `0`, but I'll have to have a look at the code first. Fired up my IDE already.

shacknetisp:
Side effects - cookzoom should only ever have meaning for weapons that both cook and zoom already; it does not determine that behaviour

MoonPadUSer:
alright, I just thought about it and came to the conclusion that it shouldn't be a problem at all when the min value of `cookzoom` is 1 since 0 would be "disabled" anyways, at least afaik, and you can just toggle of "zoom" entirely at that point. In summary, I see no problem having the min value of `cookzoom` at 1, would need some testing tho

shacknetisp:
Sub-optimal solution, will replace with checking for zero while allowing cookzoom 0 weapons.

PR #86, opened 2020-06-24
Fix cookzoom crash.

shacknetisp:
Replacement of #82 

Avoids dividing by zero in places where `cookzoom` is used which ultimately results an infinite loop.

Fixes the freeze reproduceable in master by zooming with rifle and switching to another weapon with a number key (must have `firstpersonbob 0` and no compiler optimizations and ffast-math).

### DIFF
--- a/src/game/game.cpp
+++ b/src/game/game.cpp
@@ -1010,7 +1010,7 @@ namespace game
                 int frame = lastmillis-lastzoom;
 
                 float pc = 1.f;
-                if (W(focus->weapselect, cookzoom) > 0 && frame <= W(d->weapselect, cookzoom)) {
+                if (W(d->weapselect, cookzoom) > 0 && frame <= W(d->weapselect, cookzoom)) {
                     pc = frame / float(W(d->weapselect, cookzoom));
                 }
 
@@ -2321,7 +2321,7 @@ namespace game
                 int frame = lastmillis-lastzoom;
 
                 float pc = 1.f;
-                if (W(focus->weapselect, cookzoom) > 0 && frame <= W(e->weapselect, cookzoom)) {
+                if (W(e->weapselect, cookzoom) > 0 && frame <= W(e->weapselect, cookzoom)) {
                     pc = frame / float(W(e->weapselect, cookzoom));
                 }
 
--- a/src/game/game.cpp
+++ b/src/game/game.cpp
@@ -528,9 +528,7 @@ namespace game
 
     bool inzoom()
     {
-        if(lastzoom && (zooming || lastmillis-lastzoom <= W(focus->weapselect, cookzoom)))
-            return true;
-        return false;
+        return lastzoom && (zooming || lastmillis - lastzoom <= W(focus->weapselect, cookzoom));
     }
     ICOMMAND(0, iszooming, "", (), intret(inzoom() ? 1 : 0));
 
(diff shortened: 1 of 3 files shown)


src/game/game.cpp
```diff
@@ -528,12 +528,32 @@ namespace game
...
     {
-        if(lastzoom && (zooming || lastmillis-lastzoom <= W(focus->weapselect, cookzoom)))
-            return true;
-        return false;
+        return lastzoom && (zooming || lastmillis - lastzoom <= W(focus->weapselect, cookzoom));
     }
     ICOMMAND(0, iszooming, "", (), intret(inzoom() ? 1 : 0));
 
+    // Get the progress of the zoom in the interval [0, 1], where 0 is no zoom and 1 is full zoom.
+    float zoomratio(gameent *d)
```
TheAssassin, line 536: AFAIR we decided to use snake_case for new variables and functions. Please adjust.

src/game/game.cpp
```diff
@@ -528,12 +528,32 @@ namespace game
...
+        int cookzoom = W(d->weapselect, cookzoom);
+
+        float ratio = 1.f;
+        if (cookzoom > 0 && frame <= cookzoom) {
+            ratio = frame / float(W(d->weapselect, cookzoom));
+        }
+
+        ratio = clamp(ratio, 0.f, 1.f);
+
+        if(zooming) {
```
TheAssassin, line 549: Minor style issue: we're putting braces below the statement, at least for now until we have proper code style guidelines. This follows the existing "style".

TheAssassin: I'd like to see some doc strings in that function. You should describe what `zooming` does for example (indicates whether you're zooming in or out, basically, not whether you're zooming in either direction; maybe we should rename it).

src/game/game.cpp
```diff
@@ -528,12 +528,32 @@ namespace game
...
 
+    // Get the progress of the zoom in the interval [0, 1], where 0 is no zoom and 1 is full zoom.
+    float zoomratio(gameent *d)
+    {
+        int frame = lastmillis - lastzoom;
+
+        int cookzoom = W(d->weapselect, cookzoom);
+
+        float ratio = 1.f;
+        if (cookzoom > 0 && frame <= cookzoom) {
```
TheAssassin, line 543: Let's document that `cookzoom > 0` means `cooking enabled`. Then this default value also makes a hell lot more sense (as zooming is instantaneous when cooking is not enabled for a weapon).

src/game/game.cpp
```diff
@@ -528,12 +528,32 @@ namespace game
...
+            ratio = frame / float(W(d->weapselect, cookzoom));
+        }
+
+        ratio = clamp(ratio, 0.f, 1.f);
```
TheAssassin, line 547: As written on IRC, it makes sense to document why we clamp the values at all.

src/game/game.cpp
```diff
@@ -1006,11 +1026,10 @@ namespace game
             if(d == focus && third) total *= camera1->o.dist(d->o)/(d != &player1 ? followdist : thirdpersondist);
             int prot = m_protect(gamemode, mutators), millis = d->protect(lastmillis, prot); // protect returns time left
             if(millis > 0) total *= 1.f-(float(millis)/float(prot));
+
             if(d == focus && inzoom())
             {
-                int frame = lastmillis-lastzoom;
-                float pc = frame <= W(d->weapselect, cookzoom) ? (frame)/float(W(d->weapselect, cookzoom)) : 1.f;
-                total *= zooming ? 1.f-pc : pc;
+                total *= 1.f - zoomratio(d);
```
TheAssassin, line 1032: I see... kind of non-obvious, but yeah, we're calculating some opacity here... A comment might help.

PR #87, opened 2020-06-25
Add a .editorconfig for easy editor configuration for the project

shacknetisp:
Using the format of https://editorconfig.org

Other options could be added, but indention is a big one.


TheAssassin:
It's a good start. They will have to be extended though, once we defined some proper contribution and code guidelines.

PR #89, opened 2020-06-25
Serverinfo refactoring

bmorel:
No behavior change nor bugfix are expected, this only adds documentation, and improves code readability by using RAII. Need that work for future planned features.


MoonPadUSer:
It looks beautiful

src/engine/serverbrowser.cpp
```diff
@@ -273,32 +253,35 @@ void pingservers()
...
     uchar ping[MAXTRANS];
+    // ENet is crappy enough to invert it's members depending on
+    // target OS.
+    // This does also imply that this "efficient" library wastes 25%
+    // bytes in alignment on windows for this structure.
+#ifdef WIN32
+    ENetBuffer buf = { 0, ping };
+#else
+    ENetBuffer buf = { ping, 0 };
+#endif
```
robalni, line 266: Why not do it like this, like we do in other places in the code?

```
    ENetBuffer buf;
    buf.data = ping;
    buf.dataLength = 0;
```

Then we don't need to care about the order.

TheAssassin:
```cpp
    ENetBuffer buf{};
```

Otherwise, the struct isn't initialized properly. But I'd just

```cpp
    ENetBuffer buf{0, ping};
```

src/shared/defs.hpp
```diff
@@ -0,0 +1,10 @@
+#ifndef DEFS_HPP
+#define DEFS_HPP
+
+#define SERVER_PORT 28801
+#define MAXSDESCLEN 80
+#define MAXBRANCHLEN 16
+#define MAXTRANS 5000   // max amount of data to swallow in 1 go
+#define VERSION_GAME 230
+
+#endif
```
robalni, line 10: For this file and also serverinfo.hpp: We use .h as file extension for all other header files. Maybe we should keep it the same (until we change all of them if we ever do that).

src/engine/serverbrowser.cpp
robalni: There are two problems with `ENetBuffer buf{0, ping};`:

1. We need to write different code for different operating systems as seen in this code.
2. We can't see that the right members get the right values; we have to rely on knowing the order. I like the designated initializers in C99 but if we don't have them I think the best we can do, if we want to make sure everything is initialized, is:

```
    ENetBuffer buf{};
    buf.data = ping;
    buf.dataLength = 0;
```

TheAssassin: Well, I'm not a big fan of relying on the order either, and even if it's rather unlike that things might go wrong with regards to that, it's also important to keep things consistent in the code. Some day, we might make ENet a more usable C++ library, but until this day has come, let's just keep things consistent.

We should really turn on this warning in the compiler that complains about uninitialized structs...

src/engine/serverbrowser.cpp
```diff
@@ -309,12 +292,9 @@ void checkresolver()
     int resolving = 0;
     loopv(servers)
     {
-        serverinfo &si = *servers[i];
-        if(si.resolved == serverinfo::RESOLVED) continue;
-        if(si.address.host == ENET_HOST_ANY)
+        if( servers[i]->need_resolve( resolving ) )
```
TheAssassin, line 295: From the method name and its param, I could not guess what exactly that method does. You could get away by incrementing the `resolving` variable inside this method. That's what was done before, too. I checked the implementation, and I consider it bad practice the way it's done now.

src/shared/serverinfo.cpp
```diff
@@ -0,0 +1,507 @@
...
+}
+
+bool serverinfo::need_resolve( int& resolving )
+{
+    if( m_resolved == RESOLVED || m_address.host != ENET_HOST_ANY )
+    {
+        return false;
+    }
+    int ret = m_resolved == UNRESOLVED;
+    ++resolving;
```
TheAssassin, line 413: That shouldn't be done inside this method. The method does not make any use of the variable.

src/engine/serverbrowser.cpp
```diff
@@ -309,12 +292,9 @@ void checkresolver()
...
         {
-            if(si.resolved == serverinfo::UNRESOLVED) { si.resolved = serverinfo::RESOLVING; resolverquery(si.name); }
-            resolving++;
+            resolverquery(servers[i]->name());
```
TheAssassin, line 297: Just increment `resolving` here.

src/engine/serverbrowser.cpp
```diff
@@ -309,12 +292,9 @@ void checkresolver()
...
         }
     }
     if(!resolving) return;
```
TheAssassin, line 300: Please replace by some explicit `> 0` check to increase readability.

src/engine/serverbrowser.cpp
```diff
@@ -326,18 +306,15 @@ void checkresolver()
...
-                si.resolved = serverinfo::RESOLVED;
-                si.address.host = addr.host;
                 break;
             }
         }
     }
 }
 
-static int lastreset = 0;
+int lastreset = 0;
```
TheAssassin, line 317: What's the motivation behind this change? Just curious.

src/engine/serverbrowser.cpp
```diff
@@ -218,35 +220,13 @@ int connectwithtimeout(ENetSocket sock, const char *hostname, const ENetAddress
...
-    sortedservers = false;
-
-    return si;
-}
-
 void addserver(const char *name, int port, int priority, const char *desc, const char *handle, const char *flags, const char *branch)
 {
-    loopv(servers) if(!strcmp(servers[i]->name, name) && servers[i]->port == port) return;
-    if(newserver(name, port, priority, desc, handle, flags, branch) && verbose >= 2)
+    loopv(servers) if( servers[i]->is_same( name, port ) ) return;
```
TheAssassin, line 228: I'm not a big fan of this kind of method. Maybe it's also just the naming that isn't entirely clear.

src/engine/serverbrowser.cpp
```diff
@@ -346,63 +323,34 @@ void checkpings()
...
-            getstring(text, p);
-            si->players.add(newstring(text));
-        }
-        if(gver >= 225) loopi(si->numplayers)
+        loopv(servers)
         {
-            if(p.overread()) break;
-            getstring(text, p);
-            si->handles.add(newstring(text));
+            if( servers[i]->is_same( addr ) )
```
TheAssassin, line 335: I'm really not a fan of this method. It is not a full equality check, only checks whether the address is the same. This should be reflected by the method name, too. As a reader, I initially got the impression that it advertises more than what it actually provides.

A more elegant way would be to implement a proper operator for the address. This could even be done without modifying whatever component the `addr` object is coming from (likely ENet).

src/engine/serverbrowser.cpp
```diff
@@ -346,63 +323,34 @@ void checkpings()
...
+            {
+                si = servers[i];
+                break;
+            }
         }
-        sortedservers = false;
+        if(!si && searchlan) si = serverinfo::newserver(NULL, addr.port-1, 1, NULL, NULL, NULL, NULL, addr.host);
```
TheAssassin, line 341: TODO for self: check what `newserver` does (likely some sort of factory).

TheAssassin: Please use explicit pointer checks, especially in new code. That helps increase the readability.

src/engine/serverbrowser.cpp
```diff
@@ -346,63 +323,34 @@ void checkpings()
...
+        if(!si) continue;
```
TheAssassin, line 342: Please use explicit pointer checks, especially in new code. That helps increase the readability.

src/engine/serverbrowser.cpp
```diff
@@ -346,63 +323,34 @@ void checkpings()
...
+        si->update( buf.dataLength, buf.data, serverdecay, totalmillis, lastreset );
```
TheAssassin, line 343: This `update` method is like a bundled setter, and as far as I can see only "saves" a few lines. TODO for self: check if it does anything other than setting those attributes.

src/engine/serverbrowser.cpp
```diff
@@ -526,8 +470,32 @@ void writeservercfg()
     f->printf("// servers which are connected to or queried get added here automatically\n\n");
     loopv(servers)
     {
-        serverinfo *s = servers[i];
-        f->printf("addserver %s %d %d %s %s %s %s\n", s->name, s->port, s->priority, escapestring(s->sdesc[0] ? s->sdesc : s->name), escapestring(s->authhandle), escapestring(s->flags), escapestring(s->branch));
+        servers[i]->writecfg( *f );
```
TheAssassin, line 473: Minor naming issue (this should be `write_config` or `write_cfg`, like the remaining methods).

src/engine/serverbrowser.cpp
```diff
@@ -526,8 +470,32 @@ void writeservercfg()
...
     }
     delete f;
 }
+
+VAR(IDF_PERSIST, hideincompatibleservers, 0, 0, 1);
+void getservers(int server, int prop, int idx)
```
TheAssassin, line 479: Blank line above would be appreciated.

src/engine/serverbrowser.cpp
```diff
@@ -526,8 +470,32 @@ void writeservercfg()
...
+    if(server < 0)
+    {
+        // this codes does not really hide stuff, it _removes_ stuff
+        if (hideincompatibleservers)
+        {
+            //servers.erase( std::remove_if(), servers.end()); refuses to work here, dunno why
+            //so for now let's keep the waste
+            vector<serverinfo *> servers_new;
+            std::copy_if( servers.begin(), servers.end(),
+                std::back_inserter( servers_new ),serverinfo::server_compatible );
```
TheAssassin, line 490:
```suggestion
                std::back_inserter( servers_new ), serverinfo::server_compatible );
```

src/engine/serverbrowser.cpp
```diff
@@ -526,8 +470,32 @@ void writeservercfg()
...
+            servers = servers_new;
+        }
+
+        intret(servers.length());
+    }
+    else if(servers.inrange(server))
+    {
+        servers[server]->cube_get_property( prop, idx );
```
TheAssassin, line 498: Any reason this method name is prefixed with `cube_`? TODO for myself: check corresponding header.

src/game/client.cpp
```diff
@@ -3240,142 +3240,29 @@ namespace client
         }
     }
 
-    int serverstat(serverinfo *a)
+    bool serverinfocompare(serverinfo const*a, serverinfo const*b)
```
TheAssassin, line 3243: New naming convention would suggest `serverinfo_compare`. But I do wonder why this is not implemented as an operator directly?

TheAssassin: Also, missing spaces.

```suggestion
    bool serverinfocompare(serverinfo const* a, serverinfo const* b)
```

src/game/client.cpp
```diff
@@ -3240,142 +3240,29 @@ namespace client
...
     {
-        if(a->attr.length() > 4 && a->numplayers >= a->attr[4])
+        int comp = a->version_compare( *b );
+        if( comp != 0 )
```
TheAssassin, line 3246: The motivation behind these calls is not clear. There should be additional documentary comments in the code.

src/game/client.cpp
```diff
@@ -3240,142 +3240,29 @@ namespace client
...
-                    retsw(aa->priority, ab->priority, false);
-                    break;
-                }
-                default: break;
-            }
+            comp = a->compare(*b, style, reverse);
+            if( comp != 0 ) return comp < 0;
         }
-        return strcmp(a->name, b->name);
+        return strcmp( a->name(), b->name() ) < 0;
```
TheAssassin, line 3265: Meaning changed in comparison to old code (should likely be `!= 0`, right?). If this is *really* intentional, then there should be a comment explaining why you consider all values `>= 0` here.

src/shared/defs.hpp
TheAssassin: Also, this change is not strictly related to the topic of the PR, as far as I can see (it also missed a few other definitions).

Can someone please create an issue so we can move all remaining definitions into this file properly, later on? Please don't forget to include a link to this comment.

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+#include <string>
+
+#include "defs.hpp"
+
+struct serverinfo;
+
+// list of known servers, with or without informations
+// TODO:
+//   * should be a private static of serverinfo
+//   * stop using raw pointers (RAII ftw)
```
TheAssassin, line 14: Good idea. This is the ideal use case for shared pointers.

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+#include <vector>
+#include <string>
+
+#include "defs.hpp"
+
+struct serverinfo;
+
+// list of known servers, with or without informations
+// TODO:
+//   * should be a private static of serverinfo
```
TheAssassin, line 13: Well, ideally it should not be static at all. The goal should be to no longer depend on static memory for the game's state.

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+//   * stop using raw pointers (RAII ftw)
+extern vector<serverinfo*> servers;
+
+// code imported, here to reduce build dependencies
```
TheAssassin, line 17: These are not "imports", they're forward declarations. Is that what you meant? Or do I miss something here?

There should be a TODO to remove those again before merging. They might've helped speed up your development, however they increase coupling between code entities a lot.

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+extern bool filterstring(char *dst, const char *src, bool newline, bool colour, bool whitespace, bool wsstrip, size_t len);
+namespace client
```
TheAssassin, line 19: There should be a blank line above this line, like you have one below this block.

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+}
+
+//this enum lists (some of) the server properties that can be used for sorting.
+//The C and C++ specs says and enum starts with the value 0 except if explicitly
+//specified.
+//It have 2 list of members:
+// * row 1: original names, for compat, and maybe to allow some naming support in
+//          CS latter
+// * row 2: aliases to avoid reapeating 'SINFO::' for nothing.
+// TODO: use scoped enums
```
TheAssassin, line 36: Grammar is a bit off. Also, I don't think we need to explain C++ standards there.

```suggestion
/**
 * (some of) the server properties that can be used for sorting
 *
 * it has 2 types of members:
 *  - row 1: original names, for compat, and maybe to allow some naming support in CS later
// * row 2: aliases to avoid reapeating 'SINFO::' for nothing.
// TODO: use scoped enums
```

Even though all enums are initialized with `0`, I tend to hardcode the starting value in enums by the way, following the logic "it's not a good idea if you have to explain it". I think that's also done in the remaining codebase.

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+    MM_LOCKED  ,
+    MM_PRIVATE ,
+    MM_PASSWORD,
+};
+
+struct serverinfo
+{
+    // should probably be private, but:
+    // WAITING is used in serverbrowser for cubescript exposition
+    enum
```
TheAssassin, line 83: Would it make sense to make those two values static const(expr) variables? They don't seem to have to do a lot with each other...

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+    {
+        MAXPINGS = 3,
+        WAITING = INT_MAX
+    };
+
+private:
+    // list of player/spectators names
+    std::vector<std::string> m_players;
```
TheAssassin, line 91: TODO: discuss whether to use any form of prefix. There's a lot of pros and cons against them. I've used all kinds of prefixes (and also omitted them entirely), and I have no clear preference. In my own upcoming PR I used a simple trailing underscore...

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+    // name resolution state.
+    enum
+    {
+        UNRESOLVED,
+        RESOLVING,
+        RESOLVED
+    } m_resolved;
+
+    // is "servers" sorted?
+    static bool s_sortedservers;
```
TheAssassin, line 131: Could we dump that? We can just sort before every use, it shouldn't be a very expensive operation. Keeping this kind of state appears somewhat dangerous to me.

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+    //
+    // Styles NUMPLRS and PRIO are "better if higher".
+    // Others are "better if lower".
+    // Reverse invert the sorting.
+    //
+    // returns:
+    //   * -1 if *this < other
+    //   *  0 if *this = other
+    //   *  1 if *this > other
+    int compare( serverinfo const& other, int style, bool reverse ) const;
```
TheAssassin, line 157: Do we absolutely need both overloads? I think one should just cast before use if needed.

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+    int compare( serverinfo const& other, SINFO style, bool reverse ) const;
+
+    // return 0 if versions match, -1 if *this > other and 1 otherwise
+    int version_compare( serverinfo const& other ) const;
```
TheAssassin, line 161: Method naming is pretty inconsistent. Should likely be called `compare_version`.

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+    // "return" a property of this server to CS
+    // if property < 0: returns the number of properties one can query (4)
+    // if property > 3: do nothing.
+    // if property > 0:
+    //   if index < 0: returns the size of requested array
+    //   if index > array.size(): do nothing
+    //   if property == 1: returns m_attr[index] (integer)
+    //   if property == 2: returns players[index] (result?)
+    //   if property == 3: returns handles[index] (result?)
+    // if propert == 1:
```
TheAssassin, line 178:
```suggestion
    // if property == 1:
```

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+    //   if index ==  3: returns description or name if no description (result)
+    //   if index ==  4: returns map played (result)
+    //   if index ==  5: returns number of players (integer)
+    //   if index ==  6: returns ping? Guess it's roundtrip time? (integer)
+    //   if index ==  7: returns m_lastinfo? guess it's time of last update? (integer)
+    //   if index ==  8: returns authhandle?? (result)
+    //   if index ==  9: returns flags??  (result)
+    //   if index == 10: returns branch?? (resull)
+    //   if index == 11: returns priority (integer)
+    void cube_get_property( int property, int index );
```
TheAssassin, line 191: What does `cube` refer to? The docstring implies this method is somewhat useful with regards to CubeScript. `cube` != `CubeScript`, `cube` is a game whereas CubeScript is a language. I find it a bit confusing.

Other than that, there should be a TODO to invite further contributors into reverse engineering the other values and answer the open questions.

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+
+    // write some current server's attributes into file
+    // * m_name
+    // * port
+    // * priority
+    // * description() (escaped)
+    // * authhandle (escaped)
+    // * flags (escaped)
+    // * branch (escaped)
+    void writecfg( stream& file ) const;
```
TheAssassin, line 201: Naming is a bit off, see my other comment. But it's certainly a good idea to have such a method.

*(With regards to the SRP, there should be a separate class for this kind of output, though. No action required, but it should be said once.)*

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+
+    // update server's internal informations.
+    // According to what exactly is still unknown to me.
+    // Candidate for rewrite.
+    // TODO: make private?
+    //   * only used in serverbrowser.cpp
+    void update( size_t len, void const* data, int serverdecay, int totalmillis, int lastreset );
```
TheAssassin, line 208: Should probably be provided through setters only. To the outside user, this just looks like it collapses 5 lines into one. See my other comment, too.

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+
+    // methods that allows multi-threaded server name resolution without breaking encapsulation
+    bool validate_resolve( char const* name, ENetAddress const& addr );
```
TheAssassin, line 211: There need to be proper docstrings. To me as a reader, it doesn't get clear what would be "validated" here.

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+    bool need_resolve( int& resolving );
+
+    // check if current's server network info (IP/port) match parameter(s)
+    bool is_same( ENetAddress const& addr ) const;
```
TheAssassin, line 215: As written elsewhere, the first prototype should not exist. You track the address as an instance of `ENetAddress` anyway. The more elegant approach is IMO to just implement an `operator==(const ENetAddress&, const ENetAddress&)`, and provide a getter here instead.

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+    bool is_same( char const* oname, int oport ) const;
```
TheAssassin, line 216: Not a fan of this method. I think it should be removed from the class, or be named differently to avoid giving users a wrong impression of what the method does. To me as a normal reader, it actually looks more or less like a holistic comparison operator, but it's not.

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+
+    // sends a ping to target
+    void ping( ENetSocket& sock, int serverdecay, int millis );
```
TheAssassin, line 219: Breaks the SRP to some extent.

What are the side effects of this call? This apparently likely triggers something in ENet.

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+
+    // register a new server into "servers"
+    static serverinfo *newserver(const char *name, int port = SERVER_PORT, int priority = 0, const char *desc = nullptr, const char *handle = nullptr, const char *flags = nullptr, const char *branch = nullptr, uint ip = ENET_HOST_ANY);
+
+    // check if the target is compatible with current client
+    static bool server_compatible( serverinfo const* target );
```
TheAssassin, line 225: Naming wise, this should likely be prefixed `is_`. The motivation doesn't get clear from the docstring, though. Could you please explain what "compatible" means?

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+
+    // check if current's server network info (IP/port) match parameter(s)
+    bool is_same( ENetAddress const& addr ) const;
+    bool is_same( char const* oname, int oport ) const;
+
+    // sends a ping to target
+    void ping( ENetSocket& sock, int serverdecay, int millis );
+
+    // register a new server into "servers"
+    static serverinfo *newserver(const char *name, int port = SERVER_PORT, int priority = 0, const char *desc = nullptr, const char *handle = nullptr, const char *flags = nullptr, const char *branch = nullptr, uint ip = ENET_HOST_ANY);
```
TheAssassin, line 222: Is this an implementation of the "named constructor" idiom, where it also inserts the server into the array?

src/shared/serverinfo.hpp
```diff
@@ -0,0 +1,231 @@
...
+
+    // check if the target is compatible with current client
+    static bool server_compatible( serverinfo const* target );
+
+    // sorts "servers"
+    static void sort( vector<serverinfo*> &servers );
```
TheAssassin, line 228: Do we even need the parameter? The `newserver` method also doesn't need it.

src/shared/tools.h
```diff
@@ -672,6 +673,11 @@ template <class T> struct vector
         add();
     }
 
+    void push_back( T const& el )
```
TheAssassin, line 676: I've missed that method recently in my own upcoming PR, too. In the future, please consider sending separate PRs for those.

src/shared/serverinfo.cpp
```diff
@@ -0,0 +1,507 @@
...
+    cleanup();
+}
+
+void serverinfo::reset()
+{
+    m_lastping = m_lastinfo = -1;
+    s_sortedservers = false;
+}
+
+int serverinfo::compare( serverinfo const& other, int style, bool reverse ) const
```
robalni, line 135: We usually put `const` on the left side when possible. It's confusing to mix the two styles. This is done in some other places too, not only this line.

src/shared/serverinfo.cpp
```diff
@@ -0,0 +1,507 @@
...
+		}
+		bool higher_is_better = style != SINFO_NUMPLRS && style != SINFO_PRIO;
+		if( higher_is_better != reverse ? ac < bc : ac > bc )
+		{
+			return -1;
+		}
+		if( higher_is_better != reverse ? ac > bc : ac < bc )
+		{
+			return  1;
+		}
```
robalni, line 214: Indentation is weird here.

robalni:
I can't build this. I get the error:

```
[ 30%] Building CXX object src/CMakeFiles/blue-nebula_linux.dir/engine/serverbrowser.cpp.o
/home/a/s/bluenebula/src/engine/serverbrowser.cpp: In function ‘void getservers(int, int, int)’:
/home/a/s/bluenebula/src/engine/serverbrowser.cpp:486:18: error: ‘copy_if’ is not a member of ‘std’; did you mean ‘copy’?
  486 |             std::copy_if( servers.begin(), servers.end(),
      |                  ^~~~~~~
      |                  copy
```

and if I remove that code, as already done in master, I get linker errors about undefined references to functions in shared/serverinfo.cpp and when I look I see that this file is not added to the cmake list of source files. I wonder how Travis could build it.

src/shared/serverinfo.cpp
TheAssassin: The reason is a different one. This use of `const` indicates that the method does not alter the state of the object. That means that the compiler permits its execution also for `const` references. You need to do the same for comparison operators as well. There is no way to put it on the left side.

robalni:
I have noticed that the server browser sort order jumps around every 4 seconds when I test it with these changes (if you have the mouse outside of it). It doesn't do that on master.

src/shared/serverinfo.cpp
robalni: I was talking about the `const` inside the argument list, not after the function. We would normally do `const serverinfo &other`, not `serverinfo const& other`

TheAssassin: I see. Well, in C++ it's nowadays more or less a matter of personal preference. In good old C times, the order did make a huge difference. Simple example:

```cpp
// one can modify the contents of a
char * const a;
// here, one cannot change the contents, but change the pointer address
const char* b;
```

I'm personally in preference of `const serverinfo& abc`. But we have an issue about code style open anyway. We've agreed on avoiiding such "preference style" discussions until we have a consensus (and ideally a `clang-format` file which can be used to auto-format files to some extent).

robalni: Yes, order matters if you put two things on the left side of the `const` but not if you put `const` on the left or right side of the leftmost thing.

bmorel:
> I excluded the implementation `serverinfo.cpp` from my review intentionally, but focused on the "outside view". There's a lot of questions, and a few conceptual change requests. Please don't hesitate to answer if you disagree.
> 
> Comments regarding style aren't strictly binding. It's appreciated if you'd apply a few of them, though. Feel free to discuss about them on GitHub as well. That'll help define a proper code style later on.
> 
> By the way (I could not comment those changes directly): Making the `cpp` files no longer executable is a good idea. But why also the shell scripts in `src/scripts`?

The src/scripts? Maybe some are accidents. Detail please, but this one is noisy enough. We should have a whitelist of files with the +x right anyway. And reject PRs that don't follow it without reasons, dunno how this can be done.

src/engine/serverbrowser.cpp
```diff
@@ -273,32 +253,35 @@ void pingservers()
...
+    ENetBuffer buf = { ping, 0 };
+#endif
     ucharbuf p(ping, sizeof(ping));
     putint(p, totalmillis ? totalmillis : 1);
+    buf.dataLength = p.length();
 
     static int lastping = 0;
-    if(lastping >= servers.length()) lastping = 0;
-    loopi(maxservpings ? min(servers.length(), maxservpings) : servers.length())
+    auto num_servers = servers.size();
```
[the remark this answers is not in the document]
bmorel: good point for const. For size and opt builds, I prefer to move repetitive calls into a variable. Makes debugging easier since you can inspect the variable, and shorter and thus more readable.

src/shared/serverinfo.hpp
bmorel: it sends a ping.

src/shared/tools.h
bmorel: originally, it was in a distinct commit, but yes, should have done that.

src/engine/serverbrowser.cpp
bmorel: you want "if( some_bool_value > 0 )" ?

Issue #90, opened 2020-06-25
Create Separate Weapon Variable Menu


MoonPadUSer:
I'm adding an extra menu for editing weapon variables to make the process easier, this what my earliest prototype looks like
![string](https://user-images.githubusercontent.com/37220464/85785425-920ba400-b729-11ea-8e3f-a2270dc0ba19.png)
:
when you change a variables value (so it isn't the default anymore) it'll be highlighted in orange:
![var-changed](https://user-images.githubusercontent.com/37220464/85785471-9f289300-b729-11ea-9f1a-a8e85cdb6c18.png)

The main inspiration and a big part of the code was taken from @IceflowRE, big thank you again

MoonPadUSer:
Updated version:
![new-variables-menu](https://user-images.githubusercontent.com/37220464/85861168-2fb1b280-b7c0-11ea-98ca-f0cfede4f6a4.png)
still have to do variable search and fix the alignment of the import/export buttons

MoonPadUSer:
Update:
![update-weapon-variables-menu](https://user-images.githubusercontent.com/37220464/85947554-b47d0780-b94b-11ea-8a60-14ad3f88d7f0.png)

PR #92, opened 2020-06-27
Rework variables menu

MoonPadUSer:
See issue #90 
This new menu is supposed to make it easier for newbies and oldies to modify weapons to their desire,
the file I started with was from @IceflowRE but is largely modified by me

### DIFF
--- a/config/menus/weapon_vars.cfg
+++ b/config/menus/weapon_vars.cfg
@@ -552,5 +552,5 @@ newgui weap_vars [
     ui_list [ weapmenu ]
 ]
 
-bind F9 [ showgui weap_vars ]
+bind F7 [ showgui weap_vars ]
  
--- a/config/menus/weapon_vars.cfg
+++ b/config/menus/weapon_vars.cfg
@@ -1,0 +1,556 @@
+if (getalias hideweapvarmenucategories) [] [ hideweapvarmenucategories = 0 ]
+setpersist hideweapvarmenucategories 1
+
+VAR_TYPE_INTEGER = 0
+VAR_TYPE_FLOAT   = 1
+VAR_TYPE_STRING  = 2
+
+variable_reset_button = [
+    if (var_is_modified $cur_variable_name) [ 
+        ui_button "^foReset" (format "%1 %2" $cur_variable_name $cur_variable_value_def)
+    ] [ ui_button "^faReset" ]
+]
+
+selected_export_import_action           = 0
+export_weapon_variables__all            = 0
+export_weapon_variables__server         = 0
+import_export_weapon_variables__filename = ""
+chosenvar                               = ""
+show_primary                            = 0
+
+set_chosenvar = [
+    chosenvar = $arg1
+]
+
+var_is_modified = [
+    case (getvartype $arg1) $VAR_TYPE_INTEGER [ result ( !=  (getvardef  $arg1) $$arg1 )
+    ]                       $VAR_TYPE_FLOAT   [ result ( !=  (getfvardef $arg1) $$arg1 )
+    ]                       $VAR_TYPE_STRING  [ result ( !=s (getsvardef $arg1) $$arg1 ) ]
+]
+
+vardef = [
+    var_name = (substring $arg1 (stringlen $[weap]))
+
+    ui_button (format "^f%1%2" (? (var_is_modified $arg1) "o" "w") $arg2) (format "set_chosenvar %1" $var_name) "" "radiobox" 0xFFFFFF 0xFFFFFF -1 1 (? (=s $chosenvar $var_name) "radioboxon") $ui_color_radiobutton_box
+]
+vardef2 = [
+    if $show_primary [  vardef [@[arg1]1] $arg2
+    ] [                 vardef [@[arg1]2] $arg2 ]
+]
+varcategory = [
+    if (! $hideweapvarmenucategories) [
+        ui_list [ 
+            ui_strut 2
+            ui_text (format "^fB%1" $arg1)
+        ]
+    ]
+]
+
+weapon_variables_one = [
+    [vardef  [@[weap]name]         "Name"          ]
+    [vardef  [@[weap]longname]     "Long Name"     ]
+    [vardef  [@[weap]colour]       "Colour"        ]
+    [vardef  [@[weap]cookzoom]     "Cook Zoom"     ]
+    [vardef  [@[weap]cookzoommin]  "Cook Zoom Min" ]
+    [vardef  [@[weap]cookzoommax]  "Cook Zoom Max" ]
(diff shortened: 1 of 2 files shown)


config/menus/glue.cfg
```diff
@@ -24,6 +24,20 @@ ui_right = [ ui_list [ ui_spring 1; arg1 ] ]
...
+        ui_strut $arg1
+        ui_list [
+            ui_strut $arg3
+            arg5
+            ui_strut $arg4
+        ]
+        ui_strut $arg2
+    ]
+]
+
```
shacknetisp, line 40: This needs documenting comments describing what it is and how to use it.

src/game/client.cpp
```diff
@@ -291,6 +291,99 @@ namespace client
     }
     ICOMMAND(0, writevars, "sii", (char *name, int *all, int *sv), if(!(identflags&IDF_WORLD)) writegamevars(name, *all!=0, *sv!=0));
 
+    void write_weapon_vars(const char* weapon_name, const char* file_name, bool all = false, bool server = false)
```
shacknetisp, line 294: Document what is going on in this function and what the parameters mean.

src/game/client.cpp
```diff
@@ -291,6 +291,99 @@ namespace client
...
+                    break;
+            }
+            
+        }
+        delete file_stream;
+        conoutft(CON_EVENT, "\fgSuccessfully exported \fb%s \fgconfig into file \fy%s", weapon_name, file_name);
+        
+    }
+    
+    ICOMMAND(0, writeweapvars, "ssii", (char* weapon_name, char* file_name, int* all, int* sv), if (!(identflags & IDF_WORLD)) write_weapon_vars(weapon_name, file_name, *all != 0, *sv != 0));
```
shacknetisp, line 384: This command needs description in usage.cfg

src/game/client.cpp
```diff
@@ -291,6 +291,99 @@ namespace client
...
+        for (int i = 0; i < identifiers.length(); i++)
+        {
+            ident& identifier = *identifiers[i];
+            
+            // skip if identifier is not client sided or readonly,
+            // a worldvar or doesn't start with the weapon_name
+            if ( !(identifier.flags & IDF_CLIENT)
+               || (identifier.flags & IDF_READONLY)
+               || (identifier.flags & IDF_WORLD)
+               || (strncmp(identifier.name, weapon_name, strlen(identifier.name)) == 0))
```
shacknetisp, line 328: This line is not correct. It means that the identifier will be skipped when the identifier name is equal to the weapon name. You want the identifier to be skipped when the identifier name does not begin with the weapon name.

config/menus/weapon_vars.cfg
```diff
@@ -0,0 +1,606 @@
...
+newgui weap_vars [
+    ui_header "Weapons"
+    ui_list [ weapmenu ]
+] [
+    if (= $guipasses 0) [
+        fill_variable_lists
+    ]
+]
+
+bind F7 [ showgui weap_vars ]
```
shacknetisp, line 605: This bind should be in setup.cfg with the rest of the binds.
F7 is already bound there to the team selector, though that is also bound to PERIOD, so there is some choice how to proceed.
There should also be a way to access this menu from the menu system.

config/menus/glue.cfg
MoonPadUSer: oh, yeah, ok

src/game/client.cpp
MoonPadUSer: well yes, but actually no. for some reason it saves all changed variables

config/menus/weapon_vars.cfg
MoonPadUSer: any suggestions where to put it?

shacknetisp: Probably around or in the current variables menu.

src/game/client.cpp
shacknetisp: That is because of this last test. It does not actually filter by weapon name.

config/menus/weapon_vars.cfg
MoonPadUSer: yeah, that probably makes a lot of sense, will bind the entire menu to F7 so one can easily access the other variables while they're at it too

src/game/client.cpp
MoonPadUSer: yeah I figured that out, do you have any idea how this could be achieved?

shacknetisp: One way (keeping with C string methods) is to replace
`|| (strncmp(identifier.name, weapon_name, strlen(identifier.name)) == 0)`
with 
`|| (strncmp(identifier.name, weapon_name, strlen(weapon_name)) != 0)`

Which means: continue if the identifier name does not begin with the weapon name.

src/game/client.cpp
TheAssassin: This function seems new; how about using `std::string` directly? It'll make things easier most likely.

src/game/client.cpp
MoonPadUSer: ye, I also messed up the strlen, it has to be strlen(weapon_name), not strlen(identifier.name). heh.

shacknetisp:
Some warnings to fix:

```
/home/user/src/blue-nebula/src/game/client.cpp: In function ‘void client::writegamevars(const char*, bool, bool, const char*)’:
/home/user/src/blue-nebula/src/game/client.cpp:260:45: warning: comparison with string literal results in unspecified behavior [-Waddress]
         bool export_weapon = weapon_name != "";
                                             ^~
/home/user/src/blue-nebula/src/game/client.cpp:267:30: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
         for (size_t i = 0; i < ids.size(); i++)
                            ~~^~~~~~~~~~~~
```

MoonPadUSer:
> Some warnings to fix:
> 
> ```
> /home/user/src/blue-nebula/src/game/client.cpp: In function ‘void client::writegamevars(const char*, bool, bool, const char*)’:
> /home/user/src/blue-nebula/src/game/client.cpp:260:45: warning: comparison with string literal results in unspecified behavior [-Waddress]
>          bool export_weapon = weapon_name != "";
>                                              ^~
> /home/user/src/blue-nebula/src/game/client.cpp:267:30: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
>          for (size_t i = 0; i < ids.size(); i++)
>                             ~~^~~~~~~~~~~~
> ```

on my way...

TheAssassin:
In the sense of #100, we should consider adding `-Werror -Waddress` (and maybe `-Wsign-compare`) to prevent such issues in the future.

PR #95, opened 2020-07-01
Show chatting in radar

MoonPadUSer:
Radar now shows the chat symbol when someone is chatting, I had to add the parameter `bool rotate` to `void drawblip(...) ` to prevent it from rotating tho (that's why I had to change `defend.cpp`, `capture.cpp`, etc)
It looks like this:
Screenshot from 2020-07-01 19-38-52
(I know the image quality isn't the best, but you can see it)

### DIFF
--- a/src/game/capture.cpp
+++ b/src/game/capture.cpp
@@ -87,7 +87,7 @@ namespace capture
                     }
                     else if(f.owner || f.droptime) tex = hud::alerttex;
                 }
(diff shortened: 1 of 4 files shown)


Lagetta:
Oh damn that looks so cool. Good job!

TheAssassin:
I think nobody would even percept that symbol there. It's pretty invisible...

MoonPadUSer:
It's as visible as the death symbol, as I said it's really hard to spot on the image, but it's easy to spot just because it is not the normal point texture

shacknetisp:
This is a simple feature useful in various situations (picking target based on chatting, knowing if dueling player is camping or chatting, etc.)

The size of the icon can be adjusted with `radarplayersize` and `radarplayerhintsize` and other radar variables too, however the default size should be enough to recognize the icon.

PR #97, opened 2020-07-04
Make serverbrowser search bar consistent with filtering.

shacknetisp:
Fixes the issue where the search bar would still display an inactive filter after re-opening the server browser.

Changes the placeholder text to match convention.
Adds a reset button to the filter to match other filter options.

This also prevents a clash with maps.cfg by renaming the search string to be within the serverinfo namespace and not relying on search_filter which is ALSO from maps.cfg (scope! scope! scope!)

This is an alternative PR: Closes #94

2020-07-04-090732_698x67_scrot


MoonPadUSer:
There is a reason I made #94 like that

PR #99, opened 2020-07-04
Maps list improvements: better <random> and size display.

shacknetisp:
* Fixes display of `<random>` in demos list; now displays it with the "available maps" since those are what are selected from for random.
  * Notes that `<random>` works with any mode.
* Displays the map sizes (small, medium, large) if they are set in the size variables (`smallmaps mediummaps allowmaps`)

2020-07-04-140640_472x663_scrot

### DIFF
--- a/config/menus/maps.cfg
+++ b/config/menus/maps.cfg
@@ -130,7 +130,7 @@ mapselection = [
                         ui_merge 30 [
                             if (! ($hidemappreviews)) [
                                 ui_image (? (= $mapcategory 2) $i (concatword "maps/" $i)) [] 1.5 1 "textures/emblem"
-                            ] [ 
+                            ] [
                                 ui_image "textures/emblem" [] 1.5 1
                             ]
                             ui_strut 1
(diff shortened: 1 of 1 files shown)


MoonPadUSer:
please provide screenshots

MoonPadUSer:
I would prefer if the it'd say "Random" instead of "<random>" since we're also switching from "<search>" to "Search" in searchfields

config/menus/maps.cfg
```diff
@@ -130,21 +130,58 @@ mapselection = [
...
+                                    if (!=s $map_display_sizes "") [
+                                        ui_font "little" [
+                                            ui_button $map_display_sizes []
+                                        ]
+                                    ]
+                                    ui_strut 25 (+f 1 $map_display_size_offset)
                                 ] [
                                     ui_strut 0.5
-                                    ui_button "<Random>"
+                                    ui_button "<random>"
```
MoonPadUSer, line 181: Change to "Random"

MoonPadUSer, line 181: Change to "Random"

MoonPadUSer:
If you feel like doing this: in `fillmaparray` add a check if the search starts with size: and then search for the terms after that separated by `,` e.g. `size:med,larg` should show only medium size and large maps

shacknetisp:
> If you feel like doing this: in `fillmaparray` add a check if the search starts with size: and then search for the terms after that separated by `,` e.g. `size:med,larg` should show only medium size and large maps. But I'd prefer if you/me/somebody else would do that in a separate PR

I won't include that in this PR, perhaps later; that introduces a lot of complexity.

PR #101, opened 2020-07-05
Refine Serverbrowser style

MoonPadUSer:
I adjusted the serverbrowser style so it fits more with the new map selector:
Screenshot from 2020-07-05 14-00-24
Also, the serverbrowser window doesn't resize anymore when you search for smth:
Screenshot from 2020-07-05 14-00-32
I moved the "Identify as xx" into the bottom info bar and aligned it to the right, as it takes less space like this.
The favourite star is just a placeholder for now, it doesn't do anything. I plan on adding functionality in a separate PR

### DIFF
--- a/config/menus/glue.cfg
+++ b/config/menus/glue.cfg
@@ -169,38 +169,40 @@ ui_two = [
 
 ui_big_macro = [ ui_stay_open [
     ui_list [
-        if (> $numargs 9) arg10
+        if (> $numargs 10) arg11
         ui_list [
             [@[arg1]count] = $arg2
             [@[arg1]disp] = 0
             [@[arg1]list] = 0
             ui_list [
                 ui_strut (+ $arg3 2) 1
-                ui_background $ui_color_textfield_background $ui_blend_textfield_background $ui_color_textfield_border $ui_blend_textfield_border
-                ui_strut 0.25
+                if (= $arg7 1) [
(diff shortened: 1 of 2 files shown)


config/menus/glue.cfg
```diff
@@ -169,38 +169,44 @@ ui_two = [
 
 ui_big_macro = [ ui_stay_open [
```
shacknetisp, line 170: `ui_big_macro` needs documented, especially with the changes. What is happening here and inside?

MoonPadUSer: ;_; you want me to ... understand this mess?

shacknetisp: Well, at least document what you changed.

shacknetisp:
I've played with this for a while, no issues found & it's good to have a similar style to the maps menu.

Not sure if the favorites button should be included when it's just a dummy, perhaps leave that out for now.

config/menus/glue.cfg
MoonPadUSer: sigh, alright here goes my weekend xD

MoonPadUSer:
@shacknetisp, it's just that it was rather "hard" to implement it. so I'd preferably just comment them out for now since I'll implement it very soon

MoonPadUSer:
thank you very much for completing my documentation :D

PR #104, opened 2020-07-08
Fix sound loading bug and refactor sound system a little bit

MoonPadUSer:
This PR is mainly to fix the sound bug where sounds would be offset inside the `gamesounds` vector, I fixed it by switching to a `std::map` and ended up removing the entire `addsound()` function and replacing it with `registersound()`. I later also noticed that I had to rewrite the voiceline system as it was relying on the vector based system. so I moved the entire thing into C++ too. And since I was working on the file anyways, I replaced all `char*` with `std::string` and `NULL` with `nullptr` possible without too much tinkering around in other files.


TheAssassin:
There are significant issues with this PR. First and foremost, refactoring is mixed with bug fixing. That just increases the complexity for no real reason, and makes reviewing significantly harder. Even the bugfix could've been improved in the first place.

I already started working on improving the bug fix to resolve the efficiency problems it introduces. To do so, I cherry picked the first commit of this branch, and use it as a base for further improvements. The idea of moving the "registration" of sound files in the source code was even suggested by me, and it really fixes the bug. Only the datastructures aren't quite suitable.

The idea behind this refactoring might be noble, but this would've to be done in smaller chunks, in separate PRs.

Therefore, closing for now.

PR #109, opened 2020-07-11
Add hide/mute chat functionality

MoonPadUSer:
This PR adds two checkboxes to the console options which toggle `mutechat` and `hidechat`
Screenshot from 2020-07-11 11-05-44

If `hidechat`/`mutechat` is set to 1, it only hides/mutes the chat while the player is alive,
if it is set to 2 it always hides/mutes the chat, except for hide which can be "overriden" by opening the chat textfield

### DIFF
--- a/config/menus/options.cfg
+++ b/config/menus/options.cfg
@@ -270,11 +270,14 @@ options_console = [
         ]
     ]
 
-    ui_font "little" [ ui_text "^faDefault: 32" ]
+    ui_font "little" [ ui_text "^faDefault: 32" ]; ui_strut 0.5
 
-    ui_strut 0.5
+    ui_checkbox "Play sound when mentioned in chat" playmentionedsound; ui_strut 0.5
 
-    ui_checkbox "Play sound when mentioned in chat" playmentionedsound
+    ui_list [
+        ui_checkbox "Hide Chat" hidechat; ui_strut 0.5
+        ui_checkbox "Mute Chat" mutechat
+    ]
 ]
 
 // GENERAL OPTIONS
--- a/src/game/client.cpp
+++ b/src/game/client.cpp
@@ -19,6 +19,7 @@ namespace client
     VAR(IDF_PERSIST, showservervariables, 0, 0, 1); // determines if variables set by the server are printed to the console
     VAR(IDF_PERSIST, showmapvotes, 0, 1, 3); // shows map votes, 1 = only mid-game (not intermision), 2 = at all times, 3 = verbose
     VAR(IDF_PERSIST, playmentionedsound, 0, 0, 1);
+    VAR(IDF_PERSIST, mutechat, 0, 0, 2); // mute chat sounds, 1 = mute in-game, 2 mute in-game and in spectator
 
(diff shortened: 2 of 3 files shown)


config/menus/options.cfg
```diff
@@ -270,11 +270,14 @@ options_console = [
...
+    ui_font "little" [ ui_text "^faDefault: 32" ]; ui_strut 0.5
 
-    ui_strut 0.5
+    ui_checkbox "Play sound when mentioned in chat" playmentionedsound; ui_strut 0.5
 
-    ui_checkbox "Play sound when mentioned in chat" playmentionedsound
+    ui_list [
+        ui_checkbox "Hide Chat" hidechat; ui_strut 0.5
+        ui_checkbox "Mute Chat" mutechat
+    ]
```
shacknetisp, line 280: These need an explanation of what the three-state checkboxes do.

src/game/hud.cpp
```diff
@@ -26,6 +26,8 @@ namespace hud
     FVAR(IDF_PERSIST, hudblend, 0, 1, 1);
     VAR(IDF_PERSIST, hudminimal, 0, 0, 1);
 
+    VAR(IDF_PERSIST, hidechat, 0, 0, 2); // toggle visibility of chat, 1 = hide in-game, 2 hide in-game and in spectator
```
shacknetisp, line 29: Needs a `usage.cfg` entry.

src/game/client.cpp
```diff
@@ -19,6 +19,7 @@ namespace client
     VAR(IDF_PERSIST, showservervariables, 0, 0, 1); // determines if variables set by the server are printed to the console
     VAR(IDF_PERSIST, showmapvotes, 0, 1, 3); // shows map votes, 1 = only mid-game (not intermision), 2 = at all times, 3 = verbose
     VAR(IDF_PERSIST, playmentionedsound, 0, 0, 1);
+    VAR(IDF_PERSIST, mutechat, 0, 0, 2); // mute chat sounds, 1 = mute in-game, 2 mute in-game and in spectator
```
shacknetisp, line 22: Needs a `usage.cfg` entry.

src/game/hud.cpp
shacknetisp: Also this should probably be `showchat` instead of `hidechat` to be more in line with variables like `showconsole`.

MoonPadUSer: actually I would not do that, since it's purpose is to toggle hiding the chat, not showing the chat

TheAssassin: Most variables are `show*`. Just make it `1` by default. It's more consistent.

MoonPadUSer: I do not know how I would handle that inside cubescript

config/menus/options.cfg
MoonPadUSer: added

src/game/client.cpp
MoonPadUSer: added

src/game/hud.cpp
shacknetisp: Handle in what way?

MoonPadUSer: I would prefer to keep it as "Hide chat" in the menu, but it wouldn't make sense if the variable was called `showchat`.

PR #110, opened 2020-07-11
Increase client/server update rate to 125Hz

voidanix:
Currently the update rates are at 25Hz, but it's 2020

Increase them to 125Hz

### DIFF
--- a/src/game/bomber.cpp
+++ b/src/game/bomber.cpp
@@ -626,10 +626,10 @@ namespace bomber
                 vec pos = f.pos();
                 f.distance += f.droploc.dist(pos);
                 f.droploc = pos;
-                if(f.lastowner && (f.lastowner == &game::player1 || f.lastowner->ai) && f.proj && (!f.movetime || totalmillis-f.movetime >= 40))
+                if (f.lastowner && (f.lastowner == &game::player1 || f.lastowner->ai) && f.proj && (!f.movetime || totalmillis-f.movetime >= UPDATERATE))
                 {
                     f.inertia = f.proj->vel;
-                    f.movetime = totalmillis-(totalmillis%40);
+                    f.movetime = totalmillis - (totalmillis % UPDATERATE);
                     client::addmsg(N_MOVEAFFIN, "ri8", f.lastowner->clientnum, i, int(f.droploc.x*DMF), int(f.droploc.y*DMF), int(f.droploc.z*DMF), int(f.inertia.x*DMF), int(f.inertia.y*DMF), int(f.inertia.z*DMF));
                 }
             }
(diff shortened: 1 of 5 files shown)


shacknetisp:
This should be converted to a constant instead of a magic number.

src/game/game.h
```diff
@@ -59,6 +59,10 @@ struct enttypes
     bool    noisy,  syncs,  resyncs,    syncpos,    synckin;
     const char *name,           *attrs[13];
 };
+
+// delay in milliseconds between updates for client/server
```
TheAssassin, line 63: The description is not entirely accurate. It's the minimum interval for updates. All we guarantee is that updates are not sent more often than in this interval (i.e., we guarantee a lower boundary for this delay), but we cannot ensure updates are sent at this interval (e.g., with low frame rates, updates are sent much less often).
We just increases the *possible* update rate in this PR.

src/game/bomber.cpp
```diff
@@ -626,10 +626,10 @@ namespace bomber
                 vec pos = f.pos();
                 f.distance += f.droploc.dist(pos);
                 f.droploc = pos;
-                if(f.lastowner && (f.lastowner == &game::player1 || f.lastowner->ai) && f.proj && (!f.movetime || totalmillis-f.movetime >= 40))
+                if(f.lastowner && (f.lastowner == &game::player1 || f.lastowner->ai) && f.proj && (!f.movetime || totalmillis-f.movetime >= UPDATERATE))
```
MoonPadUSer, line 629: please just add a space between `if` and `(`

src/game/capture.cpp
```diff
@@ -582,10 +582,10 @@ namespace capture
             if(f.droptime)
             {
                 f.droploc = f.pos();
-                if(f.lastowner && (f.lastowner == &game::player1 || f.lastowner->ai) && f.proj && (!f.movetime || totalmillis-f.movetime >= 40))
+                if(f.lastowner && (f.lastowner == &game::player1 || f.lastowner->ai) && f.proj && (!f.movetime || totalmillis-f.movetime >= UPDATERATE))
```
MoonPadUSer, line 585: please just add a space between `if` and `(`

src/game/client.cpp
```diff
@@ -1883,7 +1883,7 @@ namespace client
     void c2sinfo(bool force) // send update to the server
     {
         static int lastupdate = -1000;
-        if(totalmillis-lastupdate < 40 && !force) return;    // don't update faster than 25fps
+        if(totalmillis - lastupdate < UPDATERATE && !force) return;    // don't update faster than 125fps
```
MoonPadUSer, line 1886: please just add a space between `if` and `(`

PR #115, opened 2020-07-14
Rework Completion System

MoonPadUSer:
Demonstration:
[Invido](https://invidio.us/watch?v=gS3fuaDUQ5o)
[Youtube](https://www.youtube.com/watch?v=gS3fuaDUQ5o)

It shows 5 completions at once, allows switching between completions using `UP` and `DOWN`, you can complete the selected completion like usual with `TAB`

### DIFF
--- a/src/engine/console.cpp
+++ b/src/engine/console.cpp
@@ -5,6 +5,7 @@ using std::swap;
 
 #include "engine.h"
 #include "game.h"
+#include "completions.h"
 
 reversequeue<cline, MAXCONLINES> conlines;
 
@@ -472,6 +473,9 @@ bool consoleinput(const char *str, int len)
     }
     commandbuf[cmdlen + len] = '\0';
 
+    // update completion
+    completion::get_completions(commandbuf);
+
     return true;
 }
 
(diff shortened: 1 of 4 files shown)


robalni:
I tested this and here are a few problems that I found.

**Can't reach more than 5 suggestions.**

Example: I don't remember the command name "smgfragweap1" so I type "/smgfrag". With the old completion I could tab until I reached the command I wanted. With this one I can't reach it at all because I can only choose from the 5 visible suggestions.

**Arguments are deleted.**

Example: I type "/dm dark" and then press tab because I want to play on Darkness. With the old completion this works. With this one the "dark" is deleted and I only see "/dm ".

**Previous completions stay in memory and used next time.**

Example: I type "/smgfragweap1" and press tab and then escape. Now I open the chat again and type anything at all and then press tab again. Now everything I typed is deleted and replaced with "/smgfragweap1 ".

MoonPadUSer:
@robalni You are completely right, I have noticed those problems too, but sadly I haven't found the time to fix them yet :(
But to "solve" the only 5 suggestions problem I'd like to improve the completion itself because currently the order depends on the `idents` hashnameset. Thank you :)

PR #117, opened 2020-07-15
WIP: Auto spectate killer

dctremblay:


dctremblay:
remaining stuff to do ill do in 2 weeks :

- plug to make the killer spectated instead of some other random player
- second+ time killed, it is spectv instead of player spectator mode
- theres between 5000ms and 15000ms between player spect roll. must not be rolled automatically
- show deadmode keybinds when killed (home, R, scroll, space)
- make deadmode keybinds work (home, R, scroll, space)

MoonPadUSer:
why the switch to specTV when second+ time killed?

PR #120, opened 2020-07-16
Remove main menu color codes from C++

shacknetisp:
Play online, offline practice, and quit had special format code characters in text_color(), which are replaced by hex color format codes.

C++ shouldn't have such a close connection to arbitrary CubeScript.

### DIFF
--- a/config/menus/main.cfg
+++ b/config/menus/main.cfg
@@ -14,23 +14,23 @@ new_ui main [
     ] [
         ui_list [ ui_strut 30 ]
         ui_list [ ui_list [ ui_font "emphasis" [
-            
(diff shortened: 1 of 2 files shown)


TheAssassin:
A bit of a disadvantage is that when you want to change these colors, the codes are hardcoded everywhere. Any idea how to possibly solve this problem?

MoonPadUSer:
> A bit of a disadvantage is that when you want to change these colors, the codes are hardcoded everywhere. Any idea how to possibly solve this problem?

That's not a problem for this PR. those colors were only used once. that's why he removed them in the first place, I did this because I did not know `^f[0xFFFFFF]` exists

PR #132, opened 2020-07-18
Add note in log when the server cannot load a requested map.

shacknetisp:
Otherwise there is no explicit indicator on the server side what happened.

### DIFF
--- a/src/game/server.cpp
+++ b/src/game/server.cpp
@@ -3348,7 +3348,11 @@ namespace server
                 if(i == SENDMAP_MPZ) smapcrc = crcfile(reqfile);
                 mapdata[i] = openfile(reqfile, "rb");
             }
-            if(!hasmapdata()) resetmapdata();
+            if(!hasmapdata())
+            {
+                logoutf("Could not load requested map '%s' locally, will attempt to fetch it from players", reqmap);
+                resetmapdata();
+            }
         }
         copystring(smapname, reqmap);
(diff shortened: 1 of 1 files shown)


src/game/server.cpp
```diff
@@ -3348,7 +3348,11 @@ namespace server
                 if(i == SENDMAP_MPZ) smapcrc = crcfile(reqfile);
                 mapdata[i] = openfile(reqfile, "rb");
             }
-            if(!hasmapdata()) resetmapdata();
+            if(!hasmapdata())
+            {
+                logoutf("Could not load requested map: %s", reqmap);
```
TheAssassin, line 3353: Is this message shown not only when the server cannot find a map locally on the filesystem but none of the clients can upload one either?

shacknetisp: This message is shown when the server itself cannot load the map; clients may still provide it themselves.

TheAssassin: I see. Do you happen to know where we'd have to add another improved error message? Also, I'd suggest to reword this message to clearly communicate that the server couldn't find the map itself, and that it'll attempt to upload it from one of the clients and distribute it to the others.

shacknetisp: Reworded.

Adding a message for when clients haven't got the map is harder: a client's map sending can be delayed for whatever reason.

TheAssassin:
Thanks!

PR #133, opened 2020-07-18
fix Wunreachable-code-return and Wcomma

dctremblay:
its not much but its honest work

### DIFF
--- a/src/engine/irc.cpp
+++ b/src/engine/irc.cpp
@@ -360,19 +360,31 @@ ICOMMAND(0, ircbind, "ss", (const char *name, const char *s), {
 ICOMMAND(0, ircpass, "ss", (const char *name, const char *s), {
     ircnet *n = ircfind(name);
     if(!n) { conoutf("no such ircnet: %s", name); return; }
-    if(!s || !*s) { ircprintf(n, 4, NULL, "current password is: %s", n->passkey && *n->passkey ? "<set>" : "<not set>"); return; }
+    if(!s || !*s)
+    {
+        ircprintf(n, 4, NULL, "current password is: %s", *n->passkey ? "<set>" : "<not set>");
+        return;
+    }
     copystring(n->passkey, s);
 });
 ICOMMAND(0, ircauthcommand, "ss", (const char *name, const char *s), {
     ircnet *n = ircfind(name);
     if(!n) { conoutf("no such ircnet: %s", name); return; }
-    if(!s || !*s) { ircprintf(n, 4, NULL, "current auth command is: %s", n->authcommand && *n->authcommand ? "<set>" : "<not set>"); return; }
+    if(!s || !*s)
+    {
+        ircprintf(n, 4, NULL, "current auth command is: %s", *n->authcommand ? "<set>" : "<not set>");
+        return;
+    }
     copystring(n->authcommand, s);
 });
 ICOMMAND(0, ircauth, "sss", (const char *name, const char *s, const char *t), {
     ircnet *n = ircfind(name);
     if(!n) { conoutf("no such ircnet: %s", name); return; }
-    if(!s || !*s || !t || !*t) { ircprintf(n, 4, NULL, "current auth details are: %s (%s)", n->authname, n->authpass && *n->authpass ? "<set>" : "<not set>"); return; }
+    if(!s || !*s || !t || !*t)
+    {
+        ircprintf(n, 4, NULL, "current auth details are: %s (%s)", n->authname, *n->authpass ? "<set>" : "<not set>");
+        return;
+    }
     copystring(n->authname, s);
     copystring(n->authpass, t);
 });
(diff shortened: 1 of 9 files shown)


src/engine/irc.cpp
```diff
@@ -360,19 +360,31 @@ ICOMMAND(0, ircbind, "ss", (const char *name, const char *s), {
 ICOMMAND(0, ircpass, "ss", (const char *name, const char *s), {
     ircnet *n = ircfind(name);
     if(!n) { conoutf("no such ircnet: %s", name); return; }
-    if(!s || !*s) { ircprintf(n, 4, NULL, "current password is: %s", n->passkey && *n->passkey ? "<set>" : "<not set>"); return; }
+    if(!s || !*s)
```
TheAssassin, line 363: While we're at it, can you directly fix these pointer calculations to be more explicit?

```suggestion
    if(s == 0 || *s != '\0')
```

src/engine/irc.cpp
```diff
@@ -360,19 +360,31 @@ ICOMMAND(0, ircbind, "ss", (const char *name, const char *s), {
...
+    {
+        ircprintf(n, 4, NULL, "current password is: %s", *n->passkey ? "<set>" : "<not set>");
```
TheAssassin, line 365: You just reformatted here, right? Same thing in these if clauses, would appreciate if you could make pointer stuff explicit.

dctremblay: not just formating, I also removed `n->passkey && ` from `n->passkey && *n->passkey` cuz compiler said address will always return true (make sense).

src/engine/irc.cpp
dctremblay: sure ! do you want me to find all these occurences everywhere in the code and apply the same ?

TheAssassin: Can do it in this PR, if you limit it to the methods you touch anyway.

src/engine/irc.cpp
TheAssassin: Can you please post the compiler warning here completely? I'm curious.

src/engine/irc.cpp
bmorel: using `s == 0` will trigger a new warning: `zero-as-nullptr`.

dctremblay: we clean the code step after step, keeping commits as limited as possible eases the reviews.

src/engine/irc.cpp
bmorel: probably `tautological-type-limit-compare`

src/engine/rendertext.cpp
```diff
@@ -341,11 +341,11 @@ static float icon_width(const char *name, float scale)
     { \
         h++; \
         bool alt = textblinking && totalmillis%(textblinking*2) > textblinking; \
-        if(s) TEXTCOLOR(h); \
+        if(s) TEXTCOLOR(h) {} \
```
bmorel, line 344: why `TEXTCOLOR(h) {}` and not `{ TEXTCOLOR(h) }` ? (is repeated after too)

src/engine/irc.cpp
dctremblay: `address of array 'c->passkey' will always evaluate to 'true' [-Wpointer-bool-conversion]GCC`

src/engine/rendertext.cpp
dctremblay: oh *** good catch

src/engine/irc.cpp
robalni: `*s != '\0'` is not the same as `!*s`. It should be `*s == '\0'`

src/engine/irc.cpp
```diff
@@ -330,49 +330,61 @@ ICOMMAND(0, ircaddrelay, "ssisss", (const char *n, const char *s, int *p, const
 ICOMMAND(0, ircserv, "ss", (const char *name, const char *s), {
     ircnet *n = ircfind(name);
     if(!n) { conoutf("no such ircnet: %s", name); return; }
-    if(!s || !*s) { ircprintf(n, 4, NULL, "current server is: %s", n->serv); return; }
+    if(s == 0 || *s != '\0') { ircprintf(n, 4, NULL, "current server is: %s", n->serv); return; }
```
bmorel, line 333: -Wzero-as-null-pointer-constant will trigger here. Use either the inexpressive old way, or compare with nullptr. The same happens in other places.

src/engine/irc.cpp
```diff
@@ -330,49 +330,61 @@ ICOMMAND(0, ircaddrelay, "ssisss", (const char *n, const char *s, int *p, const
...
+    }
     copystring(n->authcommand, s);
 });
 ICOMMAND(0, ircauth, "sss", (const char *name, const char *s, const char *t), {
     ircnet *n = ircfind(name);
     if(!n) { conoutf("no such ircnet: %s", name); return; }
-    if(!s || !*s || !t || !*t) { ircprintf(n, 4, NULL, "current auth details are: %s (%s)", n->authname, n->authpass && *n->authpass ? "<set>" : "<not set>"); return; }
+    if(s == 0 || *s != '\0' || t == 0 || *t != '\0')
+    {
+        ircprintf(n, 4, NULL, "current auth details are: %s (%s)", n->authname, *n->authpass ? "<set>" : "<not set>");
```
bmorel, line 385: Since you edited this line, use nullptr instead of NULL otherwise -Wzero-as-null-pointer-constant will trigger here. Or maybe just keep that for a good `find -name '*.cpp' -or -name '*.h' -exec sed -i "s/\<NULL\>/nullptr/g"`. Your choice.

src/engine/irc.cpp
TheAssassin: Yep, exactly.

PR #152, opened 2020-07-25
Fix corner wall-run bug

### LINKED ISSUE #151: Collision detection bug when wallrunning across corners
This is a bug that has annoyed me for years and I think it's time to fix it.

**Summary description:** Wallruns end when you run across corners in the (-x,-y) orientation if the cube behind the corner has a visible face.

**Long description:**

Map file seen in the image: [cornerbug.zip](https://github.com/blue-nebula/base/files/4975481/cornerbug.zip)

If you try to wallrun on the yellow walls that you can see in the image (run across the corner in the middle by tapping q once, not holding it) it will work on all walls except for the one in the (-x,-y) direction; on that wall you will fall down when you reach the corner in the middle. However the wallrun does work if you cover the faces of the cube behind the wall (as done with the purple wall in the image).

**Why I think it happens:**

The reason for this is that when you reach the corner, you collide with the block behind the wall. When you run on the other walls, you actually touch the blocks behind those walls too but it only counts as collision on the (-x,-y) wall because of the order in which the game checks collision with a cube's children, negative before positive, and therefore it finds a collision with the block behind the wall before it checks the walls because the block behind has a more negative x and y coordinate than the walls themselves.

There is some code in the game that ignores collisions with cube faces that are not visible (because they are touching other cubes) so you might think that this collision should be ignored since the faces of the cube behind the wall that you are closest to are not visible. That's correct; those faces *are* ignored. The collision actually happens with the other faces of that cube (e.g. the top one). This can be demonstrated by occluding all faces of the cube like you can see at the purple wall in the background. If you do that you will see that wallruns work just fine. Another way to fix it is of course to remove the cube behind the wall.

So how is it possible to collide with e.g. the top face of the cube when you're not even close to that face? When the game decides which face you collided with, it looks for the face that you are closest to, but the face check is completely skipped for the faces that are not visible and therefore the only visible faces will be used. There is a line of code that looks like it's supposed to avoid cases like this one. The following two lines are from the definition of macro CHECKSIDE in engine/physics.cpp:

```
        if(!dir.iszero()) \
            if(d->type<ENT_CAMERA && dotval < 0 && dist < margin) continue; \
```

`continue` means that the face is skipped. `dotval` is `dir.[xyz]` or `-dir.[xyz]` depending on the face. `dir` is the velocity of the player. This line of code actually does the job; the only problem is that `dir` is sometimes 0, even when the player is moving, which will cause this line to be skipped. The reason that `dir` can be 0 is that the `collide` function is sometimes called without specifying it, like in this code line in game/physics.cpp:

```
                bool collided = collide(d);
```

The velocity (`dir`) needs to be the second argument to `collide`, otherwise it will get the default value of (0, 0, 0) and that will cause the collision check to not skip the faces it should skip. It's worth noting that the second argument was removed in commit 3027c0244ac6aa604477d1d06f319a4296e90b3e for unknown reasons.

robalni:
This fixes #151 

I don't know why the reverted commit was made but reverting it fixes the issue. I have played the game with this fix for a while now and not noticed anything different other than that the collision bug is gone.

### DIFF
--- a/src/game/physics.cpp
+++ b/src/game/physics.cpp
@@ -855,7 +855,7 @@ namespace physics
                 if(!move && !strafe) continue;
                 vecfromyawpitch(d->yaw, 0, move, strafe, dir);
                 d->o.add(dir);
-                bool collided = collide(d);
+                bool collided = collide(d, dir);
                 d->o = oldpos;
                 if(!collided || hitplayer || collidewall.iszero()) continue;
                 vec face = vec(collidewall).normalize();
@@ -875,16 +875,16 @@ namespace physics
                         if(onfloor)
                         {
                             d->o.z += space*m;
-                            if(collide(d))
+                            if(collide(d, dir))
                             {
                                 d->o.z += space*n-space*m;
-                                if(!collide(d) || hitplayer) vault = true;
+                                if(!collide(d, dir) || hitplayer) vault = true;
                             }
                         }
                         else
                         {
                             d->o.z += space*n;
-                            if(!collide(d) || hitplayer) vault = true;
+                            if(!collide(d, dir) || hitplayer) vault = true;
                         }
                         d->o = oldpos;
                     }


TheAssassin (approved):
Not an expert in that code base, but it looks good and I trust your word that it works in-game.

robalni:
I think I have found a bug caused by this change now. If you try to wall-kick right below a ceiling it sometimes fails and you don't get any speed away from the wall. This can easily be tested on testchamber right where you start because you are very close to the ceiling there. Move to the right or back wall and while facing the wall, jump and press q.

PR #153, opened 2020-07-25
remove useless macros

bmorel:
Macros only used once just makes the code harder to read (and patch)

### DIFF
--- a/src/shared/glemu.h
+++ b/src/shared/glemu.h
@@ -44,7 +44,7 @@ namespace gle
         static inline void name##f(float x) { glVertexAttrib1##f##_(index, x); } \
         static inline void name##f(float x, float y) { glVertexAttrib2##f##_(index, x, y); } \
         static inline void name##f(float x, float y, float z) { glVertexAttrib3##f##_(index, x, y, z); } \
-        static inline void name##f(float x, float y, float z, float w) { glVertexAttrib4##f##_(index, x, y, z, w); } \
+        static inline void name##f(float x, float y, float z, float w) { glVertexAttrib4f_(index, x, y, z, w); } \
         static inline void name(const vec &v) { glVertexAttrib3fv_(index, v.v); } \
         static inline void name(const vec &v, float w) { glVertexAttrib4f_(index, v.x, v.y, v.z, w); } \
         static inline void name(const vec2 &v) { glVertexAttrib2fv_(index, v.v); } \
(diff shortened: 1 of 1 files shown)


MoonPadUSer:
IMO this seems reasonable to do, but please wait for @TheAssassin to confirm

src/shared/glemu.h
```diff
@@ -40,23 +40,19 @@ namespace gle
...
-        static inline void name##suffix(type x) { glVertexAttrib1##suffix##_(index, x); } \
-        static inline void name##suffix(type x, type y) { glVertexAttrib2##suffix##_(index, x, y); } \
-        static inline void name##suffix(type x, type y, type z) { glVertexAttrib3##suffix##_(index, x, y, z); } \
-        static inline void name##suffix(type x, type y, type z, type w) { glVertexAttrib4##suffix##_(index, x, y, z, w); }
     #define GLE_INITATTRIBF(name, index) \
-        GLE_INITATTRIB(name, index, f, float) \
+        static inline void name##f(float x) { glVertexAttrib1##f##_(index, x); } \
+        static inline void name##f(float x, float y) { glVertexAttrib2##f##_(index, x, y); } \
+        static inline void name##f(float x, float y, float z) { glVertexAttrib3##f##_(index, x, y, z); } \
+        static inline void name##f(float x, float y, float z, float w) { glVertexAttrib4##f##_(index, x, y, z, w); } \
```
robalni, line 47: `f` is not a macro parameter anymore so it doesn't need to be glued using `##`.

```suggestion
        static inline void name##f(float x, float y, float z, float w) { glVertexAttrib4f_(index, x, y, z, w); } \
```

bmorel: Nice catch, thanks!

src/shared/glemu.h
```diff
@@ -44,7 +44,7 @@ namespace gle
         static inline void name##f(float x) { glVertexAttrib1##f##_(index, x); } \
         static inline void name##f(float x, float y) { glVertexAttrib2##f##_(index, x, y); } \
         static inline void name##f(float x, float y, float z) { glVertexAttrib3##f##_(index, x, y, z); } \
-        static inline void name##f(float x, float y, float z, float w) { glVertexAttrib4##f##_(index, x, y, z, w); } \
+        static inline void name##f(float x, float y, float z, float w) { glVertexAttrib4f_(index, x, y, z, w); } \
```
robalni, line 47: It was not only this line; the ones above have the same thing.

bmorel: well, turns out that most of that code is *plain useless* so I'm spending some extra time remove unneeded stuff.

bmorel:
Closing because I don't know how to change source branch

PR #155, opened 2020-07-25
Redesign profile menu

MoonPadUSer:
This redesign mainly focuses on reworking the loadout menu:
2020-07-25-165506_1920x1080_scrot
2020-07-25-165512_1920x1080_scrot
2020-07-25-165516_1920x1080_scrot
2020-07-25-165521_1920x1080_scrot
2020-07-25-165539_1920x1080_scrot
2020-07-25-165602_1920x1080_scrot

### DIFF
--- a/config/menus/profile.cfg
+++ b/config/menus/profile.cfg
@@ -1,116 +1,207 @@
 profilebuttons = [
+    local text_ok
+
+    if (&& (= $numargs 1 ) (=s $arg1 "yes_no")) [
+        text_ok = "^fgYes"
+        text_cancel = "^foNo"
+    ] [
+        text_ok = "^fgOk"
+        text_cancel = "^foCancel"
+    ]
+
     ui_list [ ui_font "emphasis" [
         ui_strut 2
-        ui_button "^fgOk" [
-            setinfo $playerprevname $playerprevcolour $playerprevmodel $playerprevvanity $playerprevloadweap $playerprevrandweap
+        ui_button $text_ok [
+            setinfo $player_name_new $player_colour_new $player_model_new $player_vanity_new $player_loadweap_new $player_randweap_new
             clear_ui 1
         ] []
         ui_strut 3
         if (needname) [
             ui_button "^foMain Menu" [show_ui main]
         ] [
-            ui_button "^foCancel" [clear_ui 1]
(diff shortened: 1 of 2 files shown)


robalni:
**Profile tab:** I think having "male" and "female" as the model options make more sense than "Default" and "Skinny", both because that is probably a more common thing to choose in games and also because the models were probably made to look like male and female bodies. Also, I don't think Charactercolor and Charactermodel should be one word.

**Loadout tab:** I like having all the weapons visible at the same time because that makes it easier and faster to change. Now you have to do an extra click to change weapons and you can't see all available weapon options at once. Another thing is that the "Show All Weapon Slots" checkbox moves when you click it.

**User Account tab:** I like the new look of this one, the way the input fields are aligned. But what happened with the checkboxes to use the account online/offline?

MoonPadUSer:
@robalni 
**Profile tab:** Using Default and Skinny instead of Male and Female is mainly to eliminate the possibility for someone to be offended by anything.
**Loadout tab:** I made it so the menu is easier to navigate for beginners and the layout is more obvious. regarding the "Show All Weapon Slots" checkbox, honestly, I don't really know how else I should do it. But I can certainly experiment a bit.
**Account tab:** after a small discussion on our IRC (#blue-nebula-dev, feel free to join) we came to a conclusion that there is no real reason for these checkboxes, but feel free to enlighten me regarding that :)

robalni:
**Profile tab:** I don't think there is any real risk to offend people. Another solution could be to just have one model because they are pretty similar to each other anyway and having different models has no gameplay purpose.

**Loadout tab:** Yes, I can see how it could be easier for a beginner to understand how it works. I still prefer the old one but I don't know how much of that is just that I'm used to it. The checkbox is not important, just a nitpick.

**Account tab:** Ok, just wanted to make sure they were useless :)

MoonPadUSer:
@robalni 
**Profile tab:** I strongly disagree with removing one model entirely as player customization is *extremely* important! Real life example would be the vanities, even though everybody knows that they do not affect gameplay in any way, they all have their own customized character :) Also, about offending people... trust me, there are 100% people who will be offended by that eventually, and just in case this game grows in popularity... I mean, it can't hurt anybody, right?

PR #157, opened 2020-07-25
glemu deadcode and macro purge

bmorel:
Each line was removed, one by one, check that the compilation passed.
On Debian 10, with clang++ and some custum setup, all passes. CI will
check that all damn systems builds. It's the only thing it's setup for,
after all.
I will *not* apply ANY codestyle or doc patch on this, because it's just
a damn cleanup, not real changes.
To redo:
* 1: apply `cpp` on your target
* 2: n = 1
* 3: until EOF
* 4: remove line 1
* 5: build
* 6: if build failed, cancel 3
* 7: n++
* 8: goto 3

### DIFF
--- a/src/shared/glemu.h
+++ b/src/shared/glemu.h
@@ -1,18 +1,19 @@
 #ifndef GLEMU_H
 #define GLEMU_H
+
 namespace gle
 {
     enum
     {
-        ATTRIB_VERTEX       = 0,
-        ATTRIB_COLOR        = 1,
-        ATTRIB_TEXCOORD0    = 2,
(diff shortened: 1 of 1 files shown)


bmorel:
FYI: I don't intend to do any code style related change on this. It's something that a tool could have do, but I know none that is able to do that automatically, so I did it (partially, cpp helped a lot) manually.
It's damn annoying, but at least, the code is now minimal, and there are no longer macros for nothing.

voidanix (approved):
If this PR is a "just remove the macros mechanically" one, then we can clean this up later with _real_ code changes.

#250 should take care of code-style changes down the road.

PR #159, opened 2020-07-25
Restore movie command

robalni:
The movie command was removed by mistake in 44b24a10332649035c9f42a91afb119738567dfe and this pull request brings it back.

### DIFF
--- a/config/glsl.cfg
+++ b/config/glsl.cfg
@@ -1853,6 +1853,133 @@ lazyshader 0 "prefab" [
     }
 ]
 
+lazyshader 0 "moviergb" [
+    attribute vec4 vvertex;
+    @(screentexcoord 0)
+    varying vec2 texcoord0;
+    void main(void)
+    {
+        gl_Position = vvertex;
+        texcoord0 = vtexcoord0;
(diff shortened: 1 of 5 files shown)


bmorel:
Could you please add some documentation around to explain what it does and how it does it, so that the mistake won't happen anew?

TheAssassin:
> was removed by mistake

It wasn't a mistake, really. IIRC we figured hardly anyone uses it. To me, at least, it was clear it was used to record, not play movies. See #23.

I'm very open to reverting those changes given there's actual users. I agree to @bmorel that it should probably be documented a bit better in order to avoid misunderstandings of any kind in the future.

TheAssassin:
I think what @bmorel meant was to add some comments in the code to mark which code belongs to the recorder to prevent future readers from thinking "ah well we don't need that any more" and removing the code...

MoonPadUSer:
> I think what @bmorel meant was to add some comments in the code to mark which code belongs to the recorder to prevent future readers from thinking "ah well we don't need that any more" and removing the code...
qwe_download

PR #160, opened 2020-07-26
Use std::swap without "using" keyword

TheAssassin:
This helps clean up the code files' headers a bit, as in a lot of headers, we've had to have "using std::swap;" in almost all code files before including other headers (mainly tool.h).

The refactoring was performed mostly by my IDE's search-and-replace feature. I've read all changes again in the diff to make sure I don't introduce issues.

### DIFF
--- a/src/engine/bih.cpp
+++ b/src/engine/bih.cpp
@@ -1,6 +1,3 @@
-#include <algorithm>
-using std::swap;
-
 #include <vector>
 #include "engine.h"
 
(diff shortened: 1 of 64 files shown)


TheAssassin:
Wasn't even hard. Just a little bit of search and replace. My IDE is great with refactoring, saves you so much time.

bmorel:
Well, it's not a problem of IDE usage vs text editor, it's more like about incremental improvements and what a mind is focused on at a time. This PR do improves things, but which one is, in your opinion, the most important change? The one that was imperfect but allowed further improvements, or this one, that only fixed more or less syntactic sugar stuff?

PR #161, opened 2020-07-26
Fix segmentation fault when changing map

robalni:
This bug was introduced in commit 9b4d1185b902e830d6eabad6d001d715c24e7ae0

The reason that the segfault happens is that when one of these loops are
used with a std::vector, the size() returns an unsigned value and `i` will
therefore be unsigned and `i>=0` will always be true.  That will cause `i`
to underflow and when used as index to an array, that's how the segfault
happens.

Steps to reproduce the bug:
Change map three times, e.g. with the following commands:
```
/dm echo
/dm cargo
/dm echo
```

### DIFF
--- a/src/shared/tools.h
+++ b/src/shared/tools.h
@@ -216,12 +216,13 @@ inline char *newstring(size_t l)                { return new char[l+1]; }
 inline char *newstring(const char *s, size_t l) { return copystring(newstring(l), s, l+1); }
 inline char *newstring(const char *s)           { if(!s) s = ""; size_t l = strlen(s); char *d = newstring(l); memcpy(d, s, l+1); return d; }
 
-#define loopv(v)     for(decltype((v).size()) i = 0; i<(v).size(); i++)
-#define loopvj(v)    for(decltype((v).size()) j = 0; j<(v).size(); j++)
-#define loopvk(v)    for(decltype((v).size()) k = 0; k<(v).size(); k++)
-#define loopvrev(v)  for(decltype((v).size()-1) i = (v).size()-1; i>=0; i--)
-#define loopvjrev(v) for(decltype((v).size()-1) j = (v).size()-1; j>=0; j--)
-#define loopvkrev(v) for(decltype((v).size()-1) k = (v).size()-1; k>=0; k--)
+// TODO: remove macros and replace them with actual for loops
+#define loopv(v)     for(decltype((v).size()) i = 0; i < (v).size(); i++)
+#define loopvj(v)    for(decltype((v).size()) j = 0; j < (v).size(); j++)
+#define loopvk(v)    for(decltype((v).size()) k = 0; k < (v).size(); k++)
+#define loopvrev(v)  for(ssize_t i = (v).size() - 1; i >= 0; i--)
+#define loopvjrev(v) for(ssize_t j = (v).size() - 1; j >= 0; j--)
+#define loopvkrev(v) for(ssize_t k = (v).size() - 1; k >= 0; k--)
 
 template <class T>
 struct databuf


bmorel:
I see your point there, but I can't reproduce. Could you tell more about your system?
I am more inclined to eradicate those horrible macros anyway.

robalni:
> I can't reproduce. Could you tell more about your system?

I use Debian unstable and g++ 9. Compiling without optimizations might also help to reproduce.

> No. There's a moment i will be == 0.

Even when i == 0, it's still true that i >= 0 and i will therefore continue to underflow.

bmorel:
Exact. Still waking up... hum. I tried the steps on my debugging branch, debian 10 + clang 7 + libc++,  maybe I should try with libstdc++...

MoonPadUSer:
I cannot reproduce the bug either Manjaro 20.04 gcc 10.1.0

voidanix:
Can reproduce on Fedora Rawhide w/ gcc 10.1.1

src/shared/tools.h
```diff
@@ -219,9 +219,9 @@ inline char *newstring(const char *s)           { if(!s) s = ""; size_t l = strl
 #define loopv(v)     for(decltype((v).size()) i = 0; i<(v).size(); i++)
 #define loopvj(v)    for(decltype((v).size()) j = 0; j<(v).size(); j++)
 #define loopvk(v)    for(decltype((v).size()) k = 0; k<(v).size(); k++)
-#define loopvrev(v)  for(decltype((v).size()-1) i = (v).size()-1; i>=0; i--)
-#define loopvjrev(v) for(decltype((v).size()-1) j = (v).size()-1; j>=0; j--)
-#define loopvkrev(v) for(decltype((v).size()-1) k = (v).size()-1; k>=0; k--)
+#define loopvrev(v)  for(int i = (v).size() - 1; i >= 0; i--)
```
TheAssassin, line 222: Well, if anything, you should probably use `ssize_t` here.

robalni: I had that idea too but I don't know how portable it is.

bmorel: It's a standard type. But the real fix would be to remove the macros and replace them with standard stuff. Since 2011 there is no more  acceptable reasons to keep those macros imo. I did so only to keep changes minimal, but they go in our way a lot too often.
We need to eradicate the `loop.*` macros. This is my opinion. 2 steps there: 1) do the job `cpp`  does and 2) use C++11 for loops when possible. 3) (yes I know) use iterators when possible.

The PR you're sending will certainly makes the current codebase more resilient (in practice and in what it looks like), but it will makes things harder for people that maintain it (more warnings about implicit conversions).

robalni: I don't think removal of those macros is relevant in this PR. That's another issue.

> more warnings about implicit conversions

I didn't see any additional warnings when I compiled.

bmorel: I have not compiled it, but, I can see additional warnings, since we are trying to integrate STL containers there.
As you said, std::vector<>::size() returns a size_t. Thus, the change would trigger new casts,while unmasking some that should have happened.
In `short: int vector<>::size()` vs `size_t std::vector<T>::size() const`. You are true, there is indeed a problem when size() returns 0 for STL. I should have spotted that.
Now, the *real* fix, that would trigger no warning, would to replace that macro so that it would use reverse iterators. I'll admit, it's my fault, I should refactor everything correctly on 1st time. I should have 1st eradicated those stupid macros, with either proper std::for_each() instances, or with for( auto& el : element_list ) and then patch the thousands of lines of code, or...
Sorry.

TheAssassin:
Just found this exact bug myself, while using "offline practice". The error occurs in https://github.com/blue-nebula/base/blob/0b8759527652ebe1eee7188364d17b873104091d/src/game/server.cpp#L3383, where the vector has only one member. `i` becomes `0xffffffff` (aka `-1`), which causes a segfault.

src/shared/tools.h
TheAssassin: I think we just use `size_t` for the forward and `ssize_t` for the reverse versions for now. Ultimately, these macros should just be replaced anyway.

PR #162, opened 2020-07-26
Add Force Vote (Veto) Button back to map selector

MoonPadUSer:
The veto button was forgotten in our new map selector menu, thus I added it back in:

first

second

third
No worries, the screenshot was just taken at a bad moment, the Force button is actually blinking like it did before

### DIFF
--- a/config/menus/maps.cfg
+++ b/config/menus/maps.cfg
@@ -776,13 +776,57 @@ newgui maps [ ui_stay_open [ octapaks [
     ui_strut 0.5
     ui_font super [
         ui_center [
+            show_force_button = 0
+            if (&& (getclientpriv $getclientnum $vetolock) (=s (getmastermode 1) "veto")) [
+                show_force_button = 1
+            ]
+
+
             if (&& (>= $chosenmode 0) (stringlen $chosenmap)) [
-                ui_button (? (isonline) "^fgVote" "^fgPlay")  [if (isonline) [showgui maps 2]; (? (= (gamemode) 1) (savewarnchk [start $chosenmap $chosenmode (chosenmuts)]) (start $chosenmap $chosenmode (chosenmuts))); mapsearch = ""]
+                ui_button (? (isonline) "^fgVote" "^fgPlay") [
+                    if (isonline) [
+                        showgui maps 2
+                    ]
+                    if (= (gamemode) 1) [
+                        savewarnchk [ start $chosenmap $chosenmode (chosenmuts)]
+                    ] [
+                        start $chosenmap $chosenmode (chosenmuts)
+                    ]
+                    mapsearch = ""
+                ]
+
+                if $show_force_button [
+                    ui_strut 1.5
+                    ui_button "^fzwyForce" [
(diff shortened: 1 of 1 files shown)


config/menus/maps.cfg
```diff
@@ -776,13 +776,57 @@ newgui maps [ ui_stay_open [ octapaks [
...
+                    mapsearch = ""
+                ]
+
+                if $show_force_button [
+                    ui_strut 1.5
+                    ui_button "^fzwyForce" [
+                        if (isonline) [
+                            showgui maps 2
+                        ]
+                        if (= (gamemode 1) [
```
shacknetisp, line 804: Missing `)`

config/menus/maps.cfg
```diff
@@ -776,13 +776,57 @@ newgui maps [ ui_stay_open [ octapaks [
...
             ]
         ]
     ]
+    ui_font "little" [ ui_center [
+        if (&& (isonline) (getclientpriv $getclientnum $masterlock) (!=s (getmastermode 1) "veto")) [
+            ui_button (format "Change mastermode from ^fg%1 ^fwto ^fzwyveto" (getmastermode 1)) [
+                mastermode (! (getmastermode))
+            ]
+        ]
+    ] ]
```
shacknetisp, line 828: This is slightly small or not spaced enough, it should be more distinguished from the button above it.

config/menus/maps.cfg
```diff
@@ -776,13 +776,57 @@ newgui maps [ ui_stay_open [ octapaks [
     ui_strut 0.5
     ui_font super [
         ui_center [
+            show_force_button = 0
+            if (&& (getclientpriv $getclientnum $vetolock) (=s (getmastermode 1) "veto")) [
+                show_force_button = 1
+            ]
```
shacknetisp, line 782: If you are showing the force button, then you probably shouldn't have the normal vote button which also forces. Perhaps have a button to force a new vote on the selection screen and a button to force your current vote on the votes screen, or a similar arrangement.

MoonPadUSer: but... the normal Vote button doesn't force?

config/menus/maps.cfg
MoonPadUSer: good catch

config/menus/maps.cfg
MoonPadUSer: huh, yeah I agree, I'll redesign that part

config/menus/maps.cfg
shacknetisp: Yes, it does, if mastermode is set to veto and you have veto privs.

MoonPadUSer: oh... well that smth we could change, Imo you should still be able to normally vote, right?

TheAssassin: No, why should you? `veto` mode is something rarely used anyway.

MoonPadUSer: the question is: why should we NOT do it

PR #163, opened 2020-07-26
Change start command to take strings

robalni:
This makes it possible to execute the start command like this:

    /start castle dm ffa duel

Most modes have an alias, like dm for deathmatch.  The old way of using
this command with numbers still works, but the difference is that the
command takes strings instead of numbers for mode and muts.  This
probably doesn't make any difference since types are implicitly
converted in cubescript.


TheAssassin:
How about (as an alternative) using the existing commands like `/dm`?

MoonPadUSer:
maybe instead call it `vote` or `suggest`since that seems more intuitive and describes its functionality better

TheAssassin:
I'm not so sure about my initial assumption any more. Looks like this new command does *not* use the voting system. But... maybe it should? I would enjoy being able to vote my favorite modes properly using the commands.

robalni:
> How about (as an alternative) using the existing commands like /dm? They could be extended to support string arguments.

They use the `start` command so making `start` support strings is a step towards making `dm` support strings.

robalni:
In these latest commits I have made it possible to use strings with `dm` and similar commands. The only problem is that you have to specify the mutators in one argument to those commands (like "ffa-basic" instead of "ffa basic") because these commands are defined as Cubescript aliases and I don't know of any way to access a list of all argument in Cubescript (like `$*` or `$@` in bash)

TheAssassin:
I'm sure @MoonPadUSer can assist with that.

MoonPadUSer:
@robalni you can get the number of arguments in cubescript by using `$numargs` IIRC

robalni:
I know how to get the number of arguments. What I tried to do was to get a list of the arguments.

MoonPadUSer:
You mean smth else than doing
```
loop i $numargs [
    $[arg@[i]]
]
```
?

robalni:
The best would be something to put instead of `$arg2` in these definitions to keep them short.

```
deathmatch      = [ start $arg1 $modeidxdeathmatch $arg2 ]
```

TheAssassin:
Keeping them short is not as much of a priority as being able to use them without `-` as separators, I would say.

robalni:
I made a command that returns a list of all arguments. I don't understand cubescript very well but I got it to work using a bit of brute force. :)

MoonPadUSer:
I just checked the cubescript code and it looks ok, no complaints from my side

src/game/client.cpp
```diff
@@ -1482,6 +1482,142 @@ namespace client
...
+        };
+
+        int nextmode = G_DEATHMATCH;
+        int nextmuts = 0;
+
+        if(mode_str && isnumeric(*mode_str))
+        {
+            nextmode = atoi(mode_str);
+        }
+        else if(mode_str && *mode_str)
```
TheAssassin, line 1541: We prefer proper pointer comparisons. Please replace with e.g.,

```suggestion
        else if(mode_str != nullptr && *mode_str != '\0')
```

or something like that.

src/game/client.cpp
```diff
@@ -1482,6 +1482,142 @@ namespace client
...
+            {"dm",         G_DEATHMATCH},
+            {"capture",    G_CAPTURE},
+            {"ctf",        G_CAPTURE},
+            {"defend",     G_DEFEND},
+            {"dac",        G_DEFEND},
+            {"bomber",     G_BOMBER},
+            {"bb",         G_BOMBER},
+            {"race",       G_RACE},
+        };
+        static const struct {const char *name; int val; int modes;} mut_names[] =
```
TheAssassin, line 1501: This should not be a struct array. Come on, we've got the power of C++ 11/14. Why not make it an `std::set<std::string, std::pair<size_t, size_t>>` or so? There is really no reason to use anonymous structs.

src/game/client.cpp
```diff
@@ -1482,6 +1482,142 @@ namespace client
...
+        int nextmuts = 0;
+
+        if(mode_str && isnumeric(*mode_str))
+        {
+            nextmode = atoi(mode_str);
+        }
+        else if(mode_str && *mode_str)
+        {
+            bool found_one = false;
+            for(auto possible_mode : mode_names)
```
TheAssassin, line 1544: Should be `const auto` or even `const auto&`.

src/game/client.cpp
```diff
@@ -1482,6 +1482,142 @@ namespace client
     }
     //ICOMMAND(0, sendmap, "", (), sendmap());
 
+    static void startcmd(char *map, char *mode_str, char *muts_str)
```
TheAssassin, line 1485: You could likely use `std::string` instead of these char arrays.

src/game/client.cpp
TheAssassin: Using a proper map will also eliminate the need to implement a custom linear search. If you wanted, you could've even used `std::find_if` here if you used an `std::vector`.

src/game/client.cpp
robalni: I usually try to follow the existing style when I write code. What are we trying to do here? Are we trying to change the style of the code? Then what's the new style that we should follow?

src/game/client.cpp
robalni: Why not use a struct array? I like it when code is simple and easy to understand. Why should I put a wrapper around my char array (called std::string) that might allocate it on the heap and do some other undefined things when I don't need that? And why should I do similar thing with the whole array? It might even be slower. I never add things to these arrays either so they don't need to grow. They are just static data.

src/game/client.cpp
robalni: I could but that doesn't give me any benefit. The only change it would make is that I would have to type ".c_str()" more inside the function.

PR #166, opened 2020-07-26
Clear server browser search bar when opened

robalni:
I suggest that we clear the search bar in the server browser when you open that menu.

Because when I search for a server and then join one, next time I open server browser I usually want to see all servers again.


MoonPadUSer:
hum, I suggest adding a checkbox in the serverbrowser options to toggle that

PR #167, opened 2020-07-27
Fix arithmetic exception when destructing particles

voidanix:
When applying the vars attached and using the flamer, when the flamer1 particles need to be destroyed (e.g. by shooting at lava), the game crashes with an arithmetic exception

Thanks to @robalni for unintentionally finding this

[segfaults.txt](https://github.com/blue-nebula/base/files/4979207/segfaults.txt)


MoonPadUSer:
does this happen with other weapons too? (would be a little weird otherwise, wouldn't it)

robalni:
I just tested it and it seems this only happens for plasma.

bmorel:
Is there is a reason here to think "f == fmax();" everytime?

voidanix:
> Is there is a reason here to think "f == fmax();" everytime?

no idea, although when replacing rnd() and rndscale() with mt19937 i noticed that the particles never fade away (and the game doesn't crash)

voidanix:
> Are there any visible consequences of this change in normal settings?

Not at all. Do note that #225 fixed this and a few other issues that involved `rnd`. Can be safely closed.

PR #171, opened 2020-07-27
Don't use fonts that don't exist

robalni:
You may have noticed that the maps menu or the graphics options suddenly change font size a few seconds after you open it while playing. This is because we were using fonts that don't exist, like "medium" and "normal". I changed them to "default".

Maybe we should handle it better when fonts don't exist (show error message?) but that's another PR.


MoonPadUSer:
Oh boy, sry for the hassle. But displaying an error message like counoutft(CON_EVENT, "\frFont \"%s\" doesn't exist", font_name); would be smth for another PR

PR #172, opened 2020-07-28
Reset slider to 0 in controls menu

robalni:
This fixes a bug where the scroll area would show too much stuff if you
scroll to the bottom of "default" and then click "spectator".  It also
makes the scroll area behave more like other ones in the game (they
reset when you change their content).

Screenshot of bug:
Screenshot

### DIFF
--- a/config/menus/options.cfg
+++ b/config/menus/options.cfg
@@ -1040,6 +1040,7 @@ editbindtitles = [
 ]
 
 curbindtype = 0
+prevbindtype = $curbindtype
 curbindshow = 1
 
 new_ui resetcontrols [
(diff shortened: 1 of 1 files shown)


MoonPadUSer:
Could you please start providing screenshots in your descriptions? makes it easier to see what you mean at a glance

robalni:
I just figured out that the reason that we see a lot of keys when we go too far down in the scroll area is that the command `searchbinds` is called with an empty string as first argument. This string can otherwise be e.g. "forward" and then the command will return the keys bound to forward (e.g. "W"). If the command gets an empty string as first argument it returns all keys and this is what happens here. The reason it gets an empty string is that we get the string from accessing a list (`@curbindacts`) with an out-of-bounds index.

With the fix done in this pull request, the index will never be out of bounds and that's how it fixes the bug.

MoonPadUSer:
I'll have to write @TheAssassin because apparently Travis CI is broken

PR #175, opened 2020-07-28
Remove loop* and loopv* macros from menus.cpp

MoonPadUSer:
Basically just doing a little bit of refactoring by getting rid of all `loop` macros in `src/engine/menus.cpp`

### DIFF
--- a/src/engine/menus.cpp
+++ b/src/engine/menus.cpp
@@ -132,8 +132,11 @@ bool pop_ui(bool skip = true)
 
 bool remove_ui(menu *m)
 {
-    loopv(menustack) if(menustack[i] == m)
+    for (int i = 0; i < menustack.length(); i++)
     {
+        if (menustack[i] != m) {
+            continue;
+        }
         menustack.remove(i);
         m->passes = 0;
         if(m->keep) *m->keep = false;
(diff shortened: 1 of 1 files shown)


robalni:
You're using a different code style, '{' on same line and space after `if` and `for`.

MoonPadUSer:
@robalni Indeed, I do. Because:
```
for (int i = 0; i < 2; i++) {
    printf("This is fine\n");
}
```

saves space and is just as easy to read as
```
for (int i = 0; i < 2; i++)
{
    printf("This is great too!\n");
}
```
I only like to do that when there's only one line inside the curly brackets (I hope I didn't violate that rule in here)

robalni:
I prefer to have everything the same, otherwise it will look like a mess after a while.

MoonPadUSer:
Hum, ok, can't hurt. :) will change it

MoonPadUSer:
alright, now it should be good.

src/engine/menus.cpp
```diff
@@ -850,7 +863,10 @@ static struct applymenu : menu
         g.start(menu_start, NULL, true);
         g.text("the following settings have changed:");
         g.pushfont("little");
-        loopv(needsapply) g.text(needsapply[i].desc, 0xFFFFFF, "point");
+        for (int i = 0; i < needsapply.length(); i++)
```
TheAssassin, line 866: I would suggest not to just stubbornly replace all occurrences of the macro, but just go for the more modern iterator-based loops where applicable. How about

```suggestion
        for (const auto& i : needsapply)
```

If you can *not* use this approach, please don't use the deprecated `length()` but the STL-compatible `size()`.

*Applies to pretty much all loops you modified.*

MoonPadUSer: Yeah, I'll do that

Issue #177, opened 2020-07-29
burndelay and shockdelay 0 causes game to crash


MoonPadUSer:
when turning `burndelay` to 0 the game will crash when you start burning.

MoonPadUSer:
The line that crashes the game seems to be
src/game/game.cpp:941
`float fluc = float(millis%burndelay)*(0.25f+0.03f)/burndelay;`

MoonPadUSer:
Just found another line that causes it too
src/game/game.cpp:3614
`else pc *= 0.75f+(float(millis%burndelay)/float(burndelay*4));`

MoonPadUSer:
another one:
src/game/hud.cpp:3289
` float pc = interval >= burntime-500 ? 1.f+(interval-(burntime-500))/500.f : (interval%burndelay)/float(burndelay/2); if(pc > 1.f) pc = 2.f-pc;`

MoonPadUSer:
I'll use std::max(burndelay, 1) instead of burndelay everywhere it's used to divide, then I'll add a comment `When playing on servers with older version, this value can be overriden to be 0, thus add a check to prevent division by 0`

robalni:
It's the same with shockdelay.

MoonPadUSer:
I'll make a separate PR for fixing shockdelay

PR #179, opened 2020-07-29
Fix game crash when burndelay is 0 bug

MoonPadUSer:
Setting the `burndelay` variable to `0` crashes the game. To fix this we can't just set `burndelay` min to `1` because servers that don't have the patch, will just override it, therefore I had to use temp vars like `_burndelay = std::max(burndelay, 1);`

Closes: #177

### DIFF
--- a/src/game/game.cpp
+++ b/src/game/game.cpp
@@ -3615,7 +3615,7 @@ namespace game
             // When playing on servers with an older version, burndelay can be 0, thus add a check to prevent division by 0
             int _burndelay = std::max(burndelay, 1);
             if(burntime - millis < _burndelay) pc *= float(burntime - millis) / _burndelay;
-            else pc *= 0.75f + (float(millis % _burndelay) / float(_burndelay * 4);
+            else pc *= 0.75f + (float(millis % _burndelay) / float(_burndelay * 4));
             vec pos = vec(d->center()).sub(vec(rnd(11)-5, rnd(11)-5, rnd(5)-2).mul(pc));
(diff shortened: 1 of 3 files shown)


voidanix:
`shockdelay` suffers from the same issue, while you are at it consider taking a look there too

MoonPadUSer:
> `shockdelay` suffers from the same issue, while you are at it consider taking a look there too

I just made PR #182 :)

src/game/game.cpp
```diff
@@ -935,10 +935,12 @@ namespace game
                     int millis = lastmillis-d->lastres[WR_BURN];
                     size_t seed = size_t(d) + (millis/50);
                     float pc = 1, amt = (millis%50)/50.0f, intensity = 0.75f+(detrnd(seed, 25)*(1-amt) + detrnd(seed + 1, 25)*amt)/100.f;
-                    if(burntime-millis < burndelay) pc *= float(burntime-millis)/float(burndelay);
+                    // When playing on servers with an older version, burndelay can be 0, thus add a check to prevent division by 0
```
TheAssassin, line 938: The comment is very long, and this is not a check you just enforce a lower boundary.

```suggestion
                    // avoid division by zero by enforcing new lower boundary
```

src/game/game.cpp
```diff
@@ -935,10 +935,12 @@ namespace game
...
+                    int _burndelay = std::max(burndelay, 1);
```
TheAssassin, line 939: The `_` prefix is preserved for standard-specified things (as per the ISO C standard) IIRC. Please rename.

MoonPadUSer: to what?

MoonPadUSer: ok, changed them to `fixed_burndelay`

PR #181, opened 2020-07-29
Replace remaining commands *gui -> *_ui

robalni:
Since we're moving from *gui commands to *_ui I thought it would be a good idea to complete that move. Doing this fixes the "TIP: press * to open this menu at any time" texts for e.g. the help menu, that would not show any key since there was a mismatch between *gui and *_ui.

Screenshot of menu not showing correct tip:
screenshot of bug


MoonPadUSer:
I think the reason we didn't change that was because it is used in inline-cubescript inside of C++ somewhere and also in custom-scripts (which we still need to write a compatibility layer for O_o)

robalni:
What is used where?

MoonPadUSer:
newgui, cleargui, etc

MoonPadUSer:
but I am not sure about that anymore, I'll have to check it

MoonPadUSer:
ok, I just looked it up, I was wrong

PR #182, opened 2020-07-30
Fix game crash when shockdelay is 0 bug

MoonPadUSer:
This PR is basically the same as #179 but for the variable `shockdelay`


TheAssassin (changes requested):
See comments in #179, they also apply here.

PR #184, opened 2020-07-31
Don't construct the string outside of vector

robalni:
This code caused the string to be constructed and then moved into the
vector.  With this change the string in constructed inside the vector.


voidanix (approved):

TheAssassin:
Replaced by #292. Thanks @robalni!

PR #185, opened 2020-08-01
Remove random AI shooting behaviour

qreeves:
Fixes "bots shoot at walls" by ensuring AI can actually see a target before shooting.


TheAssassin:
Seems to work functionality wise. However, I just had two segfaults. Investigating already.

In general, we'll first have to discuss whether this is a change that we should block until we break compat with the good ol' RE clients. We've postponed changes to the gameplay so far in order to avoid biases of any kind.

voidanix:
Superseded by #255, thanks for bringing the issue to light!

PR #187, opened 2020-08-01
Rework sounds management

TheAssassin:
This PR attempts to rewrite most of the sounds management code. It provides a new template class `soundslot_collection`, which supports both a map and vector based storage implementation, to efficiently manage map and game sounds. Also, game sounds are now registered in the C++ code so that their registration does no longer rely on matching the order of the registration calls in CubeScript with the `enum` in the code.


robalni:
Why do you use a map for `gamesounds`? I know there is a comment that says that it makes sense to do that but why? If the key is an int that comes from an enum then it feels to me like it should make more sense to use a vector.

TheAssassin:
We don't fill every slot with a sound, there are many gaps. I've first attempted to use a vector at first as well, but I experienced memory issues, and IMO the obvious approach is to use a map. We could optimize on that later on. Performance wise, you don't notice any differences.

PR #188, opened 2020-08-02
Replace custom implementation of min() and max() with std::min() and std::max()

MoonPadUSer:
This PR removes the custom implementation of `min()` and `max()` and replaces it with `std::min()` and `std::max()`

### DIFF
--- a/src/engine/animmodel.h
+++ b/src/engine/animmodel.h
@@ -40,8 +40,8 @@ struct animmodel : model
             }
             else
             {
-                fr1 = min(fr1, info.range-1)+info.frame;
-                fr2 = min(fr1+1, info.frame+info.range-1);
+                fr1 = std::min(fr1, info.range - 1) + info.frame;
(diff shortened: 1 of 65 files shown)


robalni:
Why was this done? What is the reason to use min and max from std?

MoonPadUSer:
yes, we wanna get rid of most custom written functions and types and switch to using their standard alternatives. that way we might be able to get rid of geom.h and tools.h some day

robalni:
Why do we not want to have geom.h and tools.h?

TheAssassin:
Many of the things written in `tools.h` behave in strange ways, hide memory problems or behave unexpectedly otherwise. Also, they're redundant to the STL. And they lack the necessary interface so you cannot use STL algorithms with them. They don't support modern C++ features like e.g., rvalue references. There is no point in having implemented everything by yourself. The STL is well tested and known to work, also most C++ developers know how to use it.

voidanix:
@MoonPadUSer can you please rebase and squash both this and #189?

PR #201, opened 2020-08-16
Add state for active objects to GUI

### LINKED ISSUE #200: GUI doesn't remember focus
You may have noticed that when you try to drag the scrollbar e.g. in the server browser it only works while you have the mouse cursor on it and if you release the mouse button somewhere else that counts as a click there. Or if you drag from one button to another it will count as a click on the second button.

That happens because the game doesn't store information about GUI objects to the next frame, so on the next frame the game will not know what object you were interacting with on the previous frame, and therefore it can't remember that you were using the scrollbar when the cursor leaves it.

This is annoying and it should be fixed so that you can drag-scroll while not looking at the cursor, and button clicks should only count if you pressed down the mouse button on the same GUI button as you release it on.

I'm already working on a fix for this.

robalni:
Fixes #200 about sliders that only work when the mouse is over them.

This commit adds two things; state variables (`*childpath`) with
function that modifies it, and modifications to relevant GUI object
(e.g. button, slider) functions so that they use that function to
register as a unique object and find out whether they are active.

The GUI code now keeps track of the "path" to each object.  By "path" I
mean e.g. go to the 3rd top-level child and then to it's 2nd child.
This is similar to what it looks like they do in Red Eclipse 2.  This is
not a perfect solution because it's possible that these paths change
when objects are added or removed but it is much better than just giving
each object a simple number as ID because when something is changed
inside one list it doesn't change the path to an object that is not a
child of that list.

The modifications to the GUI objects are usually done so that the
mouse-up action only happens if the object is the active object (the
mouse was pressed down on that object).

### DIFF
--- a/src/engine/ui.cpp
+++ b/src/engine/ui.cpp
@@ -95,6 +95,44 @@ struct gui : guient
     static int curdepth, curlist, xsize, ysize, curx, cury, fontdepth, mergelist, mergedepth;
     static bool hitfx, skinfx, cursorfx;
 
+    // Example: {3, 2} means that the current object (button/slider/whatever) is the 2nd child of the 3rd top-level child.
+    // The exact numbers don't matter as long as they are unique for each object.
+    static vector<int> curchildpath;
+
+    // This is a copy of curchildpath made when pressing down a mouse button on an object.
+    // It's used to remember what item is interacted with (on the previous frame).
+    static vector<int> activechildpath;
+
+    // This is used to set activechildpath on the next frame.
+    // The reason we don't set activechildpath directly is that then it will be half set in the middle of each frame.
+    static vector<int> nextactivechildpath;
+
+    bool current_is_active()
+    {
+        if(activechildpath.size() != curchildpath.size()) return false;
+        for(int i = 0; i < curchildpath.size(); i++)
+        {
+            if(curchildpath[i] != activechildpath[i]) return false;
+        }
+        return true;
+    }
+
+    // Registers current object as unique and returns whether it's active (pressed with mouse).
+    bool unique_object_active(bool hit)
+    {
+        bool active = current_is_active();
+        if(active && mouse_action[0] & GUI_UP)
+        {
+            nextactivechildpath.clear();
+        }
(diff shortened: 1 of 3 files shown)


src/engine/ui.cpp
```diff
@@ -95,6 +95,44 @@ struct gui : guient
     static int curdepth, curlist, xsize, ysize, curx, cury, fontdepth, mergelist, mergedepth;
     static bool hitfx, skinfx, cursorfx;
 
+    // Example: {3, 2} means that the current object (button/slider/whatever) is the 2nd child of the 3rd top-level child.
+    // The exact numbers don't matter as long as they are unique for each object.
+    static vector<int> curchildpath;
```
MoonPadUSer, line 100: why not use `std::vector`?

MoonPadUSer, line 100: why not use `std::vector `?

src/engine/ui.cpp
```diff
@@ -313,10 +353,13 @@ struct gui : guient
             mergelist = curlist;
             mergedepth = curdepth;
         }
+        if(!curchildpath.empty()) curchildpath.last()++;
+        curchildpath.emplace_back(0);
     }
 
     int poplist()
     {
+        if(curchildpath.size()) curchildpath.pop_back();
```
MoonPadUSer, line 362: please write `if (int(curchildpath.size()) != 0)` instead of `if (curchildpath.size())`

MoonPadUSer:
functionality wise it looks pretty good so far

src/engine/ui.cpp
robalni: Because vector is used in other places. It could be confusing to mix the two.

MoonPadUSer: We wanna switch to using STL datatypes and functions instead of the custom implementation, IIRC the custom vector has severe memory problems, but also it's way easier to use the STL for everybody because everybody usually knows how it behaves, if you look at PRs that have been merged so far, they all use `std::vector` instead of `vector` unless it has a serious advantage to use `vector`

PR #207, opened 2020-09-06
Skip SDL if not building client

robalni:
This makes it possible to compile the server on computers that don't have SDL2.


TheAssassin (approved):

TheAssassin:
Good catch. Thanks!

PR #209, opened 2020-09-10
Crosshair separation

robalni:
The mouse cursor and the crosshair are currently treated as the same thing. The same variables are used to store their state and the same functions are used to render them. This was unexpected to me when I found it while trying to make changes to the mouse cursor.

In this pull request they are two different things. This makes it possible to e.g. show the crosshair and cursor in different positions at the same time and therefore it's also possible to keep the crosshair in the background while showing menus.

### DIFF
--- a/src/engine/rendergl.cpp
+++ b/src/engine/rendergl.cpp
@@ -1855,7 +1855,6 @@ void drawfadedslice(float start, float length, float x, float y, float size, flo
 
 float cursorx = 0.5f, cursory = 0.5f;
 float crosshairx = 0.5f, crosshairy = 0.5f;
-vec cursordir(0, 0, 0);
 vec crosshairdir(0, 0, 0);
 
 struct framebuffercopy
--- a/src/engine/rendergl.cpp
+++ b/src/engine/rendergl.cpp
@@ -1854,7 +1854,9 @@ void drawfadedslice(float start, float length, float x, float y, float size, flo
 }
 
 float cursorx = 0.5f, cursory = 0.5f;
+float crosshairx = 0.5f, crosshairy = 0.5f;
 vec cursordir(0, 0, 0);
+vec crosshairdir(0, 0, 0);
 
 struct framebuffercopy
 {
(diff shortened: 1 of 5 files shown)


src/game/hud.cpp
```diff
@@ -1310,12 +1310,54 @@ namespace hud
...
+    static int choose_pointer_type()
     {
-        int size = int((pointertype == POINTER_GUI ? cursorsize : crosshairsize)*hudsize);
-        float fade = pointertype == POINTER_GUI ? cursorblend : crosshairblend;
+        return hasinput() && !(!hasinput(true) || commandmillis > 0)
+            ? POINTER_GUI
+            : POINTER_NONE;
+    }
+
+    static void draw_pointer(int w, int h, int pointertype)
```
MoonPadUSer, line 1320: would be great if you could use `ivec2 dims` instead of` int w, int h`

src/engine/rendergl.cpp
```diff
@@ -1854,7 +1854,8 @@ void drawfadedslice(float start, float length, float x, float y, float size, flo
 }
 
 float cursorx = 0.5f, cursory = 0.5f;
-vec cursordir(0, 0, 0);
+float crosshairx = 0.5f, crosshairy = 0.5f;
```
MoonPadUSer, line 1857: why not use `vec2 crosshair_pos `instead of two floats?

robalni: Because I wanted to make it similar to cursorx and cursory to make the code more consistent.

MoonPadUSer: why not make `float cursorx` and `cursory` into a `vec2 cursor_pos` ?

robalni: Because that was not part of what I changed. You have to limit it somewhere.

MoonPadUSer: ok c:

MoonPadUSer (changes requested):
Looks and works great

PR #210, opened 2020-09-10
Simplify skin functions

robalni:
This change makes the code easier to read. It changes behaviour of drawskin slightly; now it stretches the image instead of repeating it but that should not be visible with default skin.

### DIFF
--- a/src/engine/rendergl.cpp
+++ b/src/engine/rendergl.cpp
@@ -2337,21 +2337,29 @@ void cleanupgl()
 
 void drawskin(Texture *t, int x1, int y1, int x2, int y2, int colour, float blend, int size, const matrix4x3 *m)
 {
-    int w = max(x2-x1, 2), h = max(y2-y1, 2), tw = size ? size : t->w, th = size ? size : t->h;
-    float pw = tw*0.25f, ph = th*0.25f, qw = tw*0.5f, qh = th*0.5f, px = 0, py = 0, tx = 0, ty = 0;
-    if(w < qw)
+    int width = max(x2 - x1, 2);
+    int height = max(y2 - y1, 2);
+    int tw = size ? size : t->w;
+    int th = size ? size : t->h;
+    int corner_w = tw * 0.25f;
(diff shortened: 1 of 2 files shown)


src/engine/rendergl.cpp
```diff
@@ -2337,21 +2337,29 @@ void cleanupgl()
 
 void drawskin(Texture *t, int x1, int y1, int x2, int y2, int colour, float blend, int size, const matrix4x3 *m)
 {
-    int w = max(x2-x1, 2), h = max(y2-y1, 2), tw = size ? size : t->w, th = size ? size : t->h;
-    float pw = tw*0.25f, ph = th*0.25f, qw = tw*0.5f, qh = th*0.5f, px = 0, py = 0, tx = 0, ty = 0;
-    if(w < qw)
+    int width = max(x2 - x1, 2);
```
MoonPadUSer, line 2340: it'd be awesome if you could change `max` and `min` to `std::max` and `std::min` c:

src/engine/rendergl.cpp
```diff
@@ -2337,21 +2337,29 @@ void cleanupgl()
...
+    int height = max(y2 - y1, 2);
+    int tw = size ? size : t->w;
```
MoonPadUSer, line 2342: please write `size != 0 ? ...` instead of just `size ? ...`

MoonPadUSer:
Isn't the drawskin() function supposed to use some kind of 9-split and repeat it based on that? Instead of just plain stretching/repeating

robalni:
Yes, it still splits it in 9. It's just the center piece that is stretched now instead of repeated.

MoonPadUSer:
I think that's not how you're supposed to do that, ideally it repeats it so it can give ideal results

robalni:
Doesn't that just depend on what skin you have? Some skins look better stretched and some repeated.

MoonPadUSer:
yeah, but usually you design your skins in a way so they look the best when repeated, e.g. if you have a low-res skin, stretched will make it look very pixelated

PR #213, opened 2020-09-12
Add multiline comments to cubescript

robalni:
This pull request adds `/* this type of comment */` to cubescript, requested by @MoonPadUSer.

### DIFF
--- a/src/engine/command.cpp
+++ b/src/engine/command.cpp
@@ -1028,8 +1028,32 @@ static inline void skipcomments(const char *&p)
     for(;;)
     {
         p += strspn(p, " \t\r");
-        if(p[0]!='/' || p[1]!='/') break;
-        p += strcspn(p, "\n\0");
+        if(p[0] == '/' && p[1] == '/')
+        {
+            p += strcspn(p, "\n\0");
+        }
+        else if(p[0] == '/' && p[1] == '*')
+        {
+            p += 2;
+            for(;;)
+            {
+                p += strcspn(p, "*\0");
+                if(p[0] == '*' && p[1] == '/')
+                {
+                    p += 2;
+                    break;
+                }
+                else if(p[0] == '\0')
+                {
+                    break;
+                }
+                p++;
+            }
+        }
+        else
+        {
+            break;
+        }
     }
 }
 
--- a/src/engine/command.cpp
+++ b/src/engine/command.cpp
@@ -1027,11 +1027,16 @@ static inline void skipcomments(const char *&p)
 {
     for(;;)
     {
+        // Skip whitespace before comments.
         p += strspn(p, " \t\r");
+
+        // Skip single-line comment if found.
         if(p[0] == '/' && p[1] == '/')
         {
             p += strcspn(p, "\n\0");
         }
+
+        // Skip multi-line comment if found.
         else if(p[0] == '/' && p[1] == '*')
         {
             p += 2;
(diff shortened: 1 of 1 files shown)


MoonPadUSer:
It'd be great if you could add some comments if possible, since you seem to understand the compiler quite a bit

MoonPadUSer:
running this:
```
echo "testacle" /*
echo "test"
*/ echo "lala"
```
results in
`testacle echo lala`
which is theoretically not a big deal but I think it'd be nice if 
the `*/` is on another line than `/*` if it'd act like a `;`, that way you don't combine parts that aren't meant to be combined

robalni:
Logically the behaviour you see there should be the correct one because you commented out the statement terminator (newline) and therefore the statements should become the same statement. I can also see why it should not be the correct behaviour because in a line-based language, statements on different lines should be different statements. This makes me feel like multiline comments might not be a good idea in line-based languages because it's not obvious what the correct behaviour should be. The main problem is that newline characters are invisible and that you can't put the comment-start in the end of the line after the newline character, otherwise it would be obvious what should happen. I don't know what the best thing to do here is.

MoonPadUSer:
I mean, it wouldn't be that big of a deal to leave it as-is tbh

src/engine/command.cpp
```diff
@@ -1028,8 +1028,32 @@ static inline void skipcomments(const char *&p)
     for(;;)
     {
         p += strspn(p, " \t\r");
-        if(p[0]!='/' || p[1]!='/') break;
-        p += strcspn(p, "\n\0");
+        if(p[0] == '/' && p[1] == '/')
+        {
+            p += strcspn(p, "\n\0");
```
TheAssassin, line 1033: Never heard of `strcspn` ever before, but it's quite useful indeed. Nice.

src/engine/command.cpp
```diff
@@ -1028,8 +1028,32 @@ static inline void skipcomments(const char *&p)
...
+        }
+        else if(p[0] == '/' && p[1] == '*')
+        {
+            p += 2;
+            for(;;)
+            {
+                p += strcspn(p, "*\0");
```
TheAssassin, line 1040: I'm confused. How can a null byte (which I think `\0` represents) be useful here? The second parameter is already a null-terminated string. The array equivalent of that string should be `['*', 0, 0]` as far as I can see... Does this have any effect at all? If yes, I think you'll have to document it in a comment.

src/engine/command.cpp
```diff
@@ -1028,8 +1028,32 @@ static inline void skipcomments(const char *&p)
     for(;;)
     {
         p += strspn(p, " \t\r");
-        if(p[0]!='/' || p[1]!='/') break;
-        p += strcspn(p, "\n\0");
+        if(p[0] == '/' && p[1] == '/')
```
TheAssassin, line 1031: I think we should document that this block is supposed to handle single-line comments, like the others handle multi-line ones. Also, a blank line above this one will increase readability.

TheAssassin (changes requested):
Looks pretty clean, although I wish there were a few more comments. I've got a few questions and a few suggestions.

src/engine/command.cpp
robalni: Haha, I didn't think about that. Good catch. The old code was doing it like this too.

TheAssassin (approved):
Haven't got to testing this, but I'll take @robalni's and @MoonPadUSer's word for it. Looks good!

PR #214, opened 2020-09-13
Add Colorscheme Options

MoonPadUSer:
This PR adds the option to choose between 2 colorschemes, one of them being the default red eclipse colorscheme and the other being an experimental blue nebula colorscheme I created, this is mainly to provide an easy way for users to try out colorschemes and give feedback
2020-09-13-204904_1680x1050_scrot

### DIFF
--- a/config/colorschemes.cfg
+++ b/config/colorschemes.cfg
@@ -1,0 +1,55 @@
+colorscheme_default = [
+    resetvar ui_blend_background
+    resetvar ui_blend_slider
+    resetvar ui_blend_slider_border
+    resetvar ui_blend_textfield_background
+    resetvar ui_blend_textfield_border
+    resetvar ui_color_active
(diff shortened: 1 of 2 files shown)


voidanix (changes requested):
1. This does not work apparently, seems to fail with a `unknown command: setcolors` in the console when pushing the 2 new buttons.
2. The `active` button and the `pick a colour for the variable /ui_color_active` seem clunky and confusing.
3. Got some whitespace at https://github.com/blue-nebula/base/pull/214/files#diff-59c8b041b4a3f9cd0dfa9d5311b1a7994a540dff82848bb05970571c99ee7f7eR306
4. The `colorscheme_` "containers" (?) should probably have a docstring e.g. why is the default one resetting vars?

Barely knowing any cubescript _at all_, this is only a general review but hopefully you get the idea.

PR #216, opened 2020-10-04
Remove variable `frameloops`

robalni:
It was hard to understand what it meant and it was only used for
determining whether the loop was in the first frame/iteration.
Therefore I replaced it with variable `first_frame` which has a clearer
purpose.


MoonPadUSer (approved):
Looks good to me, please wait for another approval tho

TheAssassin:
The purpose of the new variable is indeed clearer. However, I ask myself, can't we just get rid of these `if`s completely? Perhaps by moving the stuff outside the `if`s before the loop (risking some code duplication, but I could imagine not all calls might be needed in the first frame anyway)?

voidanix:
I think the question should really be _what is the **actual** purpose of `frameloops` aka `first_frame`_?

Removing the variable with its if-statements altogether results in the exact same behavior; plus, considering the `int frames` parameter does absolutely nothing, it could very likely be a remnant from old code.

If this can be dropped without regressions, then better take that route.

voidanix (approved):

voidanix:
Superseded by #291

PR #218, opened 2020-10-11
Simplify draw functions

robalni:
This makes the flow of the code more direct and linear and easy to follow. The code should still do exactly the same thing as before.


TheAssassin (approved):
Can be refactored further in a future PR.

PR #220, opened 2020-11-16
Fix vanity radiobutton issue

robalni:
This fixes an issue where the radiobuttons on the right side were not selected correctly when clicking one on the left side; they would always be on "nothing" even if you have something.


TheAssassin (approved):
Looks good to me. @MoonPadUSer please check the PR, too, and merge it if it looks good to you, too.

PR #221, opened 2020-12-04
Fix building on OpenBSD

TheAssassin:
Thanks to @voidanix for testing the if clause.


MoonPadUSer (approved):

voidanix (approved):
LGTM, tested on f34 and OBSD 6.8

TheAssassin:
Now Travis CI doesn't let me restart the build (claims I'm not in any organization, which is clearly not true; been happening ever since they announced their weird plans on moving to the `.com` version). Hence I'm changing the settings so PRs are not blocked by Travis any more.

Only macOS failed, and that one only because of some timeout. I'll merge it anyway.

PR #223, opened 2020-12-05
GitHub actions

### LINKED ISSUE #222: CI, yet again
Given that https://travis-ci.org is going to close down in just a few weeks, we need to move on to another platform.

It looks like GitHub actions would work. They even provide macOS build environments, which we need in order to be able to provide binaries for this (very very annoying) OS.

Looking for volunteers! It shouldn't be too much work, since I wrote most of the CI scripts independent from Travis (except for a few minor things maybe). I'm available for any questions. I'd like this to be tested in a fork of this main repository first, before we then move everything over to a new system.

TheAssassin:
Fixes #222.

This PR replaces Travis CI with GitHub Actions, which is not only faster, but also nicer to use. Plus it doesn't suffer from Travis CI's business decisions, which basically killed FOSS builds on their platform.


mhmdanas (approved):

.github/workflows/main.yml
```diff
@@ -0,0 +1,52 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  build-nix:
+    name: Ubuntu xenial (x86_64) debug
+    runs-on: ubuntu-16.04
+    env:
+      BUILD_TYPE: Debug
```
voidanix, line 10: Consider using `BUILD_TYPE: Release` as a global env and skip the Debug here, it's pretty much useless

.github/workflows/main.yml
```diff
@@ -0,0 +1,52 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        run: sudo apt-get update && sudo apt-get install -y libsdl2-dev libsdl2-image-dev libsdl2-mixer-dev zlib1g-dev
+      - name: Build for nix
+        run: bash -xe ci/build-nix.sh
+
+  build-nix-release:
+    name: Ubuntu xenial (x86_64) release
+    runs-on: ubuntu-16.04
```
voidanix, line 20: Xenial is going EOL in a few months, consider using bionic

voidanix (changes requested):
Few nitpicks and please squash the first 3 commits

.github/workflows/main.yml
TheAssassin: We're probably going to target `xenial` as long as it's not completely EOL with our AppImage builds, so I'd rather keep things consistent here.

.github/workflows/main.yml
TheAssassin: I'm not an expert with regards to GitHub actions. Mind to push the necessary changes?

TheAssassin: Well, actually, it makes sense to build things not just with the release config on the CI, but also with `Debug` as a target. Most of us use the latter locally for development, and the former is used in our release builds.

TheAssassin:
Seems to build now, @voidanix. I fixed your last commit by moving `env` to the top level, where it belongs according to GitHub's docs.

voidanix (approved):
LGTM now, sorry about the fuss

PR #227, opened 2020-12-24
Remove X11 as build dependency, remove numlock detection, rework capslock detection

voidanix:
Makes caps lock actually be recognized under wayland, needs testing on Windows and macOS

### DIFF
--- a/src/engine/console.cpp
+++ b/src/engine/console.cpp
@@ -628,9 +628,6 @@ void processkey(int code, bool isdown)
         case SDLK_CAPSLOCK:
             if(!isdown) capslockon = capslocked();
             break;
-        case SDLK_NUMLOCKCLEAR:
-            if(!isdown) numlockon = numlocked();
-            break;
         default: break;
     }
     keym *haskey = keyms.access(code);
@@ -942,10 +939,9 @@ void writecompletions(stream *f)
 
 #ifdef __APPLE__
 extern bool mac_capslock();
-extern bool mac_numlock();
 #endif
 
-bool capslockon = false, numlockon = false;
+bool capslockon = false;
 #if !defined(WIN32) && !defined(__APPLE__)
 #include <X11/XKBlib.h>
 #endif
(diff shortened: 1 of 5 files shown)


TheAssassin:
Quite a bunch of changes. I hope X11 still works, though?

voidanix:
> Quite a bunch of changes. I hope X11 still works, though?

Works on GNOME X11, should work for other DEs/WMs too

src/engine/console.cpp
```diff
@@ -940,55 +937,17 @@ void writecompletions(stream *f)
...
-    if(d)
-    {
-        uint n = 0;
-        XkbGetIndicatorState(d, XkbUseCoreKbd, &n);
-        XCloseDisplay(d);
-        return (n&0x01)!=0;
-    }
-    #endif
-    return false;
+    if((SDL_GetModState() & KMOD_CAPS) == KMOD_CAPS)
```
TheAssassin, line 943: You could also just check whether the result is non-zero. But this works just fine as well.

Why not a bit easier, though?

```cpp
    return (SDL_GetModState() & KMOD_CAPS) == KMOD_CAPS;
```

The return value is `bool`, not `int` anyway...

TheAssassin (commented):
Looks good to me mostly, just have one suggestion where you can save 3 LOC.

src/engine/console.cpp
```diff
@@ -940,55 +937,14 @@ void writecompletions(stream *f)
...
-    {
-        uint n = 0;
-        XkbGetIndicatorState(d, XkbUseCoreKbd, &n);
-        XCloseDisplay(d);
-        return (n&0x01)!=0;
-    }
-    #endif
-    return false;
+    // returns 1 if capslock is on
+    return (SDL_GetModState() & KMOD_CAPS) == KMOD_CAPS;
```
TheAssassin, line 944: Well, it's actually `true` that is returned. But I guess people get the idea.

TheAssassin (approved):

TheAssassin:
@voidanix please fix the linking issue on Windows, then we can merge this PR.

PR #230, opened 2021-05-23
Switch to Libera.Chat

TheAssassin:
We decided to move away from freenode in the wake of the recent events. This commit updates relevant docs as well as the in-game tips and the built-in IRC chat client to use the new channel on Libera.Chat.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ The builds are, as of yet, **fully compatible with Red Eclipse 1.5.8 to 1.6.0**.
 
 ## Get in touch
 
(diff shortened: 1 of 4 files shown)


robalni:
There is also README.md that mentions freenode.

TheAssassin:
Good catch. Updated the README, too.

voidanix (approved):
There's also [this](https://github.com/blue-nebula/base/blob/911303c8b28ac34ad3c6dae644675cfa6c76ce43/src/engine/irc.cpp#L246) where you can remove the workaround altogether; so far LGTM

TheAssassin:
Well, I'm not sure the client code won't be removed or at least reworked anyway sooner or later. Didn't want to touch that comment because it might still be valid, and I don't have time to check that.

PR #233, opened 2021-07-20
Split teamkilling moderation for bots and players

### LINKED ISSUE #186: seperating /teamkillban with 2 new vars: /botteamkillban and /playerteamkillban
There is a command called `teamkillban` that gives warnings before person is banned from the server. I think that this command should be seperated to players and bots. There are some servers like openvar ones where killing alliy bots is almost inevitable with very strong weapons. While bots don't matter when they are killed, players might get really annoyed by it (myself included). You kill no matter a bot or a player, only that it is an ally, you could still get a ban which is kinda stupid.

voidanix:
This PR introduces a new variable name `teamkillsplitactortype` which moderates team killing based on the actor type of the target.

There are also six new variables to moderate between the players and bots:
```
playerteamkillwarn
playerteamkillkick
playerteamkillban

botteamkillwarn
botteamkillkick
botteamkillban
```

I have tested this on a private server with only bots and it seems to work correctly.

This PR will not break compatibility with existing 1.6.0 servers.

Fixes #186


TheAssassin:
> Unsure if this breaks compatibility with existing 1.6.0 servers.

This will alter gameplay to some extent, but in a positive way that also applies to existing clients.

I'm concerned that the variables will require a protocol break, though. I'm not entirely sure, but I think that variables are implicitly assigned an index (IIRC some enum will be generated using these macros or something like that), and that they are transmitted to the clients by those indexes. Old clients might not be able to deal with the new variables and might assign the wrong values to the variables they know.

voidanix:
It seems that changing variables on a server is completely server-side: if the variable is not supported on the server, then an `unknown variable` message will show up.

If it is not supported by the client then they can still execute/modify that variable because the server provides it (it autocompletes too).

tl;dr this will **not** break compatibility with existing servers, I will update the PR body to reflect that.

TheAssassin:
That sounds great, thank you. I'll try to validate your results tonight, too, to further make sure that we don't run into compatibility issues.

A detailed code review will follow afterwards.

src/game/vars.h
```diff
@@ -89,12 +89,21 @@ GVAR(IDF_ADMIN, floodtime, 250, 10000, VAR_MAX); // time span to check for flood
 GVAR(IDF_ADMIN, floodlines, 1, 5, VAR_MAX); // number of lines in aforementioned span before too many
 
 GVAR(IDF_ADMIN, teamkilllock, 0, PRIV_OPERATOR, PRIV_MAX);
+GVAR(IDF_ADMIN, teamkillsplitactortype, 0, 0, 3); // pick what sort of teamkills to count/punish (0: bots and players, 1: bots and players separately, 2: players only, 3: bots only)
```
TheAssassin, line 92: I think `0` is unnecessary and confusing. I would strictly follow the respective config for bots and players.

How about a binary scheme, though?

```
0x0 (0b00) -> none
0x1 (0b01) -> players only
0x2 (0b10) -> bots only
0x3 (0b11) -> both bots and players
```

This is done a lot within the game.

TheAssassin: In fact, I'm not sure we need the variable at all. If you don't want to ban for bots, just set all three other variables to 0, right? This variable is therefore kind of redundant.

voidanix: > I think `0` is unnecessary and confusing. I would strictly follow the respective config for bots and players.

What do you exactly mean by this?

`0` keeps the current behaviour, `1` counts teamkills for each actor type (e.g. teamkilling three times with `0` gets you a warn, but with `1` you can teamkill 2 bots and 2 players before getting a warn).

> How about a binary scheme, though?
> 
> ```
> 0x0 (0b00) -> none
> 0x1 (0b01) -> players only
> 0x2 (0b10) -> bots only
> 0x3 (0b11) -> both bots and players
> ```
> 
> This is done a lot within the game.

Well, it's not a bad idea but I don't see the point for it; if I understood you correctly, `none` would be very confusing because it is essentially `teamkilllock`.

> If you don't want to ban for bots, just set all three other variables to 0, right? This variable is therefore kind of redundant.

With `teamkillsplitactortype 0` setting the `botteamkill*` variables to `0` would do nothing because the game will count every teamkill regardless of the actor type; in the same scenario but with `teamkillsplitactortype {1, 3}` you would actually be warned/kicked/banned at the first teamkill.

This is because the `*teamkillkick` and `*teamkillban` variables operate on the amount of warnings the player received.

Essentially you have set `0 frags until first warning, 0 warns until kick/ban`.

voidanix: If you don't want to punish for bots just do `teamkillsplitactortype 2`.

An alternative but worse way to do this would be setting `teamkillsplitactortype 1` and `botteamkillban` to VAR_MAX, just remember that you will be kicked and warned.

TheAssassin: > With teamkillsplitactortype 0 setting the botteamkill* variables to 0 would do nothing because the game will count every teamkill regardless of the actor type

That's essentially what I find confusing. It's not clear (to an end user) which counter will be incremented when the "old" behavior is enabled. Then, the bot variables are not used, but bots are counted as players. The variable name does not indicate that.

Perhaps we're thinking too complex here to begin with... This PR adds 7 new variables `teamkillsplitactortype`. That's quite a lot.

voidanix: > It's not clear (to an end user) which counter will be incremented when the "old" behavior is enabled. Then, the bot variables are not used, but bots are counted as players. The variable name does not indicate that.

What do you propose then? Renaming `teamkillsplitactortype` to `teamkillpunishtype`?

If you were thinking about removing the `playerteamkill*` and `botteamkill*` variables in favour of only keeping the variable mentioned above (which would lead `teamkillsplitactortype 1` to not exist), then there could be some control problems.

For example if I would like on a server to ban for teamkilling bots after 10 warnings, just using `teamkillsplitactortype 2` is going to leave killing bots completely unmoderated (which might be a good thing for plasma2 users, but not for the rest of the team).

TheAssassin: I think you could implement "unmoderated" by setting these variables to `0` (or maybe `-1`, although I don't see a point in an additional value of 0, really). That's a pattern used a lot in other variables. And it would eliminate two of the four `teamkillsplitactortype` values. And the variable name then also makes more sense, because then you just need to define whether to merge both counters or not, i.e., have a binary variable: when set to 0, only the `playerteamkill*` variables are respected and bots are counted as players, when it's not 0 they're counted separately. Seems pretty straightforward to me, no reason to remember the meanings of the different values.

If we represent "unmoderated" by setting the counters to a negative, we could further move the control variable into the `botteamkill*` "namespace" to give the user an idea that if it is set, these variables become useful. This is also done for other variable types within the game already. How about something like `botteamkillenable` or `botteamkillcountseparately`? (These are just some ideas, we should try to come up with something better).

TheAssassin (approved):
Feel free to merge, @voidanix.

TheAssassin:
Thank you!

PR #240, opened 2021-08-18
Add warn command

voidanix:
Before this commit, warnings sent by moderators were not a real thing:
they were simple messages with some cool blinking yellow.

Add an actual warn command that respects the syntax of existing moderation
commands, that is `/command cn reason`.

By default only moderators can warn players; this can be modified with
the new `warnlock` variable.


src/game/client.cpp
```diff
@@ -1147,8 +1147,16 @@ namespace client
             defformatstring(st, " (to team %s)", game::colourteam(f->team));
             concatstring(name, st);
         }
-        if(flags&SAY_ACTION) formatstring(line, "\fv* %s %s", name, msg);
-        else formatstring(line, "\fw<%s> %s", name, msg);
+        else if (flags&SAY_WARNING)
+        {
+            defformatstring(swarn, "%s\fzwy, ", game::colourname(t));
+            prependstring(msg, swarn);
```
voidanix, line 1153: This would have been way easier with `std::string`, we need to find time for that someday.

TheAssassin (approved):

PR #242, opened 2021-09-09
Add enet as submodule

voidanix:
Also upgrades its version to 1.3.17


.github/workflows/main.yml
```diff
@@ -15,6 +15,8 @@ jobs:
       - uses: actions/checkout@v2
       - name: Install dependencies
         run: sudo apt-get install -y libsdl2-dev libsdl2-image-dev libsdl2-mixer-dev zlib1g-dev
+      - name: Checkout enet submodule
```
TheAssassin, line 18: You should do this before installing the dependencies IMO. I like grouping tasks of the same kind, helps keeping track of things.
It's a bit of a shame that the `checkout` action doesn't provide a feature to select the submodules you want to check out...

TheAssassin (approved):
Thanks!

PR #244, opened 2021-09-14
Add percentage for capture/defend CTF countdowns

### LINKED ISSUE #234: adding ctf protect count downs
While playing capture the flag protect mode while holding the flag, you can't clearly see how long you and the enemy are holding the flag to get a point. Even if it's shown it is not visible enough.

I am thinking if it's possible to place it in these logos, or do it like in bomber a count down.
image

voidanix:
Fixes #234

Sadly the code for printing notices is too messy, but a `Flag capture in: X seconds` for protect CTF would have been nice

### DIFF
--- a/src/game/capture.cpp
+++ b/src/game/capture.cpp
@@ -217,14 +217,6 @@ namespace capture
                 }
                 else if(millis <= 1000) skew += (1.f-skew)-(clamp(float(millis)/1000.f, 0.f, 1.f)*(1.f-skew));
                 int oldy = y-sy;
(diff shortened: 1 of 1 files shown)


TheAssassin:
Could you please provide a screenshot?

voidanix:
Screenshot

TheAssassin (approved):

TheAssassin:
I also think this resolves #234. Perhaps in the future we'll be able to also add something more apparent like in regular bb or bb-hold.

TheAssassin:
Thanks @voidanix.

PR #248, opened 2022-06-17
Fix some warnings and general cleanup

voidanix:
- Initialize x, y, z inside vec
- Remove volatile keyword from errors counter
- Clear objects of now trivial type
- Fix possible uninitialization of height
- Convert weapon selection system to std::vector
- Fix -Wnon-pod-varargs error on clang 14
- Fix -Wint-in-bool-context
- Document what the sound enums actually are
- Fix unneeded parenthesis
- Fix uninitialized values in md3 structs
- Fix comparison between different signedness
- Fix address of array always returning true
- Remove set but unused variables
- Fix implicit conversion between float and int
- Fix shift with negative value
- Attempt to fix misleading indentation warning


src/engine/renderva.cpp
```diff
@@ -1338,11 +1339,10 @@ void renderzpass(renderstate &cur, vtxarray *va)
     if(cur.vbuf!=va->vbuf) changevbuf(cur, RENDERPASS_Z, va);
     if(!cur.depthmask) { cur.depthmask = true; glDepthMask(GL_TRUE); }
     if(cur.colormask) { cur.colormask = false; glColorMask(GL_FALSE, GL_FALSE, GL_FALSE, GL_FALSE); }
-    int firsttex = 0, numtris = va->tris;
+    int numtris = va->tris;
```
TheAssassin, line 1342: Whenever you change such declarations, please consider using `auto`.

src/game/ai.cpp
```diff
@@ -146,7 +146,7 @@ namespace ai
 
     int weappref(gameent *d)
     {
-        if(d->loadweap.length()) return d->loadweap[0];
+        if(d->loadweap.size()) return d->loadweap[0];
```
TheAssassin, line 149: I really appreciate the use of explicit comparisons. This implicitly casts the resulting numeric value to `bool`. This form is just cumbersome to read.

~~Also, isn't there an `empty()` function? It'd be a much more idiomatic approach to this. We should add one if there is none.~~ This was changed to an STL container, so there clearly is an `empty` function which you should use here.

src/game/client.cpp
```diff
@@ -465,7 +465,7 @@ namespace client
 
     void setloadweap(const char *list)
     {
-        vector<int> items;
+        std::vector<int> items;
```
TheAssassin, line 468: What's the motivation for this change?

src/game/client.cpp
```diff
@@ -477,11 +477,14 @@ namespace client
                 items.emplace_back(v >= W_OFFSET && v < W_ITEM ? v : 0);
             }
         }
-        game::player1.loadweap.shrink(0);
-        loopv(items) if(game::player1.loadweap.find(items[i]) < 0)
-        {
-            game::player1.loadweap.add(items[i]);
-            if(game::player1.loadweap.length() >= W_LOADOUT) break;
+        game::player1.loadweap.clear();
+        for (const auto &i : items) {
```
TheAssassin, line 481: The use of reference adds overhead for integral numeric types, as `i` needs to be dereferenced before use, and the pointer consumes additional memory (an overhead of 8 bytes on a 64-bit system compared to the typical 4 bytes for an `int`). It is more efficient to copy here.

src/game/client.cpp
```diff
@@ -477,11 +477,14 @@ namespace client
...
+            if (std::find(game::player1.loadweap.begin(),
```
TheAssassin, line 482: Why do you have to use `std::end` over `loadweap.end()`?

src/game/client.cpp
```diff
@@ -477,11 +477,14 @@ namespace client
...
+                game::player1.loadweap.end(), i) == std::end(game::player1.loadweap)) // NOT found
+            {
+                game::player1.loadweap.emplace_back(i);
+                if(game::player1.loadweap.size() >= W_LOADOUT) break;
```
TheAssassin, line 486: Please avoid the use of same-line conditional instructions when touching code. Moving the code to another line (ideally with brackets enclosing the scope) signalizes the branching. I usually also insert a blank line above `if`s etc.

src/shared/tools.h
```diff
@@ -140,6 +141,16 @@ static inline int bitscan(uint mask)
 #define PRINTFARGS(fmt, args)
 #endif
 
+template<class T>
+const bool inrange(std::vector<T> vec, size_t i) {
+    return i < vec.size();
```
TheAssassin, line 146: Do these functions really add any useful functionality? I don't think so, honestly.

src/shared/tools.h
```diff
@@ -140,6 +141,16 @@ static inline int bitscan(uint mask)
...
+}
+
+template<class T>
+const bool inrange(std::vector<T> vec, int i) {
+    return i >= 0 && i < static_cast<int>(vec.size());
```
TheAssassin, line 151: You shouldn't need two specializations. Just define one generic template (maybe `static_assert`ing the use of a numeric type for the second template parameter) and cast to `decltype(vec.size())`.

In any case, narrowing `vec.size()` to `int` makes no sense, you should always cast to the larger size.

src/shared/tools.cpp
```diff
@@ -95,7 +95,7 @@ int getuint(ucharbuf &p)
         n += (p.get() << 7) - 0x80;
         if(n & (1<<14)) n += (p.get() << 14) - (1<<14);
         if(n & (1<<21)) n += (p.get() << 21) - (1<<21);
-        if(n & (1<<28)) n |= -1<<28;
+        if(n & (1<<28)) n |= 0xF0000000;
```
TheAssassin, line 98: I don't see the motivation behind this change.

src/shared/geom.h
```diff
@@ -77,7 +77,7 @@ struct vec
         float v[3];
     };
 
-    vec() {}
+    vec() : x(0), y(0), z(0) {}
```
TheAssassin, line 80: You should rather initialize all three directly next to their declaration instead of using the constructor. This way, if there ever is the need for additional constructors, these values will have a sane default.

It is usually a good idea to do this for all integral types (e.g., numeric types, pointers (`= nullptr`), etc.

src/shared/ents.h
```diff
@@ -114,13 +114,12 @@ struct physent : baseent                        // can be affected by physics
     uchar type;                                 // one of ENT_* above
     uchar collidetype;                          // one of COLLIDE_* above
 
-    physent() : speed(100), weight(100), radius(3), aboveeye(1),
-        xradius(3), yradius(3), zradius(14), zmargin(0), curscale(1), speedscale(1),
+    physent() : speed(100), weight(100), radius(3), height(14), aboveeye(1),
```
TheAssassin, line 117: While we're at it, how about moving the initialization of all these numeric values from the constructor to the variable declaration? Also, I think this makes the initialization of `zradius` from `height` more robust, as the compiler will not just warn but error out if the order is mixed up (if `height` ever was declared first, its initial (uninitialized random) value will not be assigned to `zradius`, since it is unknown to the compiler at that stage).

src/game/server.cpp
```diff
@@ -3384,7 +3384,7 @@ namespace server
                 {
                     prev.erase( prev.begin() + i );
                 }
-                while(prev.size() >= G(maphistory))
+                while(prev.size() >= (size_t)(G(maphistory)))
```
TheAssassin, line 3387: Please use C++ casts. You may also just use `decltype` to infer the target type from `size` (but that's totally overkill here, since `size`'s type is well defined and won't ever change unless the class changes, and even then, I'd guarantee it'll never be a narrowing cast).

src/game/physics.cpp
```diff
@@ -1055,7 +1055,7 @@ namespace physics
             if(curmat == MAT_WATER || oldmat == MAT_WATER)
             {
                 const bvec &watercol = getwatercol((curmat == MAT_WATER ? matid : pl->inmaterial) & MATF_INDEX);
-                mattrig(bottom, watercol, 0.5f, int(radius), PHYSMILLIS, 0.25f, PART_SPARK, curmat != MAT_WATER ? S_SPLASH2 : S_SPLASH1);
+                mattrig(bottom, watercol, 0.5f, int(radius), PHYSMILLIS, 0.25f, PART_SPARK, (curmat != MAT_WATER ? S_SPLASH2 : S_SPLASH1));
```
TheAssassin, line 1058: Just a readability change, right?

src/game/game.h
```diff
@@ -552,17 +552,18 @@ struct clientstate
...
 
     clientstate() : colour(0), model(0), checkpointspawn(1), weapselect(W_CLAW), lastdeath(0), lastspawn(0), lastpain(0), lastregen(0), lastregenamt(0), lastbuff(0), lastshoot(0),
         actortype(A_PLAYER), spawnpoint(-1), ownernum(-1), skill(0), points(0), frags(0), deaths(0), totalpoints(0), totalfrags(0), totaldeaths(0), spree(0), lasttimeplayed(0), timeplayed(0),
         cpmillis(0), cptime(0), queuepos(-1), quarantine(false)
     {
         vanity[0] = '\0';
-        loadweap.shrink(0);
-        lastweap.shrink(0);
-        randweap.shrink(0);
+        loadweap.clear();
```
TheAssassin, line 563: There is no reason for these calls in the constructor. All of them will be initialized empty, since `std::vector` has a default constructor.

src/game/game.h
```diff
@@ -552,17 +552,18 @@ struct clientstate
     int actortype, spawnpoint, ownernum, skill, points, frags, deaths, totalpoints, totalfrags, totaldeaths, spree, lasttimeplayed, timeplayed, cpmillis, cptime, queuepos;
     bool quarantine;
     string vanity;
-    vector<int> loadweap, lastweap, randweap;
+    std::vector<int> loadweap, lastweap, randweap;
```
TheAssassin, line 555: Again, is there a specific reason to change these? I'd just like to know your thinking.

TheAssassin (changes requested):
Had a first look. Mostly great improvements of the code. I have a few remarks I'd like to discuss, though.

src/engine/main.cpp
```diff
@@ -115,7 +115,7 @@ void quit()                  // normal exit
     exit(EXIT_SUCCESS);
 }
 
-volatile int errors = 0;
```
robalni, line 118: I think the `volatile` keyword was used because the `fatal` function can be called from signal handlers, which the compiler can't predict, so we always need to get the latest value of `errors` from memory.

src/game/aiman.h
```diff
@@ -214,7 +214,7 @@ namespace aiman
         else if(ci->aireinit >= 1)
         {
             if(ci->aireinit == 2) loopk(W_MAX) loopj(2) ci->weapshots[k][j].reset();
-            sendf(-1, 1, "ri6si3siv", N_INITAI, ci->clientnum, ci->ownernum, ci->actortype, ci->spawnpoint, ci->skill, ci->name, ci->team, ci->colour, ci->model, ci->vanity, ci->loadweap.length(), ci->loadweap.length(), ci->loadweap.getbuf());
+            sendf(-1, 1, "ri6si3siv", N_INITAI, ci->clientnum, ci->ownernum, ci->actortype, ci->spawnpoint, ci->skill, ci->name, ci->team, ci->colour, ci->model, ci->vanity, ci->loadweap.size(), ci->loadweap.size(), &ci->loadweap);
```
robalni, line 217: On the change from `ci->loadweap.getbuf()` to `&ci->loadweap`, have you tested that this works? When I look in the `sendf` function, it seems to assume that the second argument to the `v` format specifier is an `int` pointer to the internal buffer, not to the `vector` object.

src/shared/tools.cpp
robalni: Just like the commit message says, I think this is to remove a warning about shifting a negative value. Another way to solve this that would keep the shift is: `-1u<<28` or `~0u<<28`.

src/game/aiman.h
TheAssassin: Good catch. You have to pass `ci->loadweap.data()`.

src/engine/irc.cpp
```diff
@@ -357,19 +357,19 @@ ICOMMAND(0, ircbind, "ss", (const char *name, const char *s), {
 ICOMMAND(0, ircpass, "ss", (const char *name, const char *s), {
     ircnet *n = ircfind(name);
     if(!n) { conoutf("no such ircnet: %s", name); return; }
-    if(!s || !*s) { ircprintf(n, 4, NULL, "current password is: %s", n->passkey && *n->passkey ? "<set>" : "<not set>"); return; }
+    if(!s || !*s) { ircprintf(n, 4, NULL, "current password is: %s", *n->passkey ? "<set>" : "<not set>"); return; }
```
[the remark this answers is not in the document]
voidanix: Because the address of an array always returns true when compared against something? Even if the array is empty, that's what it will give.

Consider the following toy code:
```c
#include <stdio.h>
#include <stdbool.h>

int main() {
	int hello[5];
	bool aa = true;

	if (hello && aa)
		puts("I'm true!");
	else
		puts("I'm not!");

	printf("%d\n", hello);

	return 0;
}
```
Even if you aggregate-initialize that `hello[5]`, it will always be true and the compiler rightfully says so as well.

src/engine/renderva.cpp
voidanix: Could you be a bit more specific? `va->tris` is an `int` here, could not find any reason to `auto` this unless it was some long/annoying type to write over and over (and `numtris` seems kind of self-explanatory)

src/game/ai.cpp
voidanix: True that, solving

src/game/client.cpp
voidanix: IIRC there was a compiler warning in the `sendf` from aiman.h reviewed by @robalni related to the custom vector's `getbuf()`, so here we are

src/game/client.cpp
voidanix: True that, solving

src/game/client.cpp
voidanix: No idea what I thought of when writing that...

I'm also guessing it is better to use `std::any_of`?

src/game/client.cpp
voidanix: Agreed, will fix formatting

src/shared/tools.h
voidanix: It helps replacing the existing custom vector with std::vector though

src/shared/tools.h
voidanix: True that, will try to solve

src/shared/geom.h
voidanix: Sure

src/shared/ents.h
voidanix: Wanted to do this but thought of just keeping the code the way it was, sure

src/game/server.cpp
voidanix: Already done in current last commit

src/game/physics.cpp
voidanix: Kind of: the compiler is probably interpreting this as `MAT_WATER ? S_SPLASH2 : S_SPLASH1` and then comparing against `curmat`

src/game/game.h
voidanix: True that, solving

src/engine/main.cpp
voidanix: But all `fatal` does is show up to 1 error window (more specifically `SDL_MESSAGEBOX_ERROR`) and write up to 2 fatal errors in log.txt, correct?

If so, what could be done is either:
* Show only 1 error in the log (the one printed by the window)
* Use `std::atomic<unsigned short>` or something to make the amount of fatal errors configurable for the log

src/game/aiman.h
voidanix: True that, didn't notice the `int*`

src/game/physics.cpp
robalni: != is computed before ?:

src/engine/main.cpp
robalni: I can't see any current risks with removing `volatile` since the code inside `fatal` calls functions defined in other translation units while `errors` is not static, so the compiler can't make any assumptions about the access to `errors`, assuming link time optimization is not used, so `errors` is probably both written to and read from memory anyway. But I still think it's a good idea in general to make global or static variables accessed by signal handlers volatile just to be sure the compiler doesn't do anything unwanted.

voidanix:
Thanks for the reviews @TheAssassin and @robalni ! Updated the branch

src/engine/renderva.cpp
TheAssassin: This is just a general suggestion I had while reading through your code. `auto` just makes sure no narrowing can ever happen in the future. But I guess `-Wall` helps us avoid this problem.

src/shared/ents.h
TheAssassin: Valid point.

src/game/physics.cpp
TheAssassin: @robalni is right here. If you want to make this more explicit (and improve readability), you have to put brackets around the `!=` comparison.

src/engine/main.cpp
TheAssassin: Can we just get rid of this variable entirely?

robalni: The problem with getting rid of the variable is that then we can get recursive errors if a signal happens while the signal handler is executing.

src/game/physics.cpp
voidanix: That does not work I'm afraid:
```
bn/src/game/physics.cpp: In function ‘void physics::updatematerial(physent*, const vec&, const vec&, bool)’:
bn/src/game/physics.cpp:1058:115: warning: ‘?:’ using integer constants in boolean context, the expression will always evaluate to ‘true’ [-Wint-in-bool-context]
 1058 |                 mattrig(bottom, watercol, 0.5f, int(radius), PHYSMILLIS, 0.25f, PART_SPARK, (curmat != MAT_WATER) ? S_SPLASH2 : S_SPLASH1);
      |                                                                                                                   ^
```

robalni: I think the problem here is that in the definition of the macro `mattrig`, the last argument `mw` is used like:

```c
if(mw >= 0)
```

and this expands to:

```c
if(curmat != MAT_WATER ? S_SPLASH2 : S_SPLASH1 >= 0)
```

which is the same as:

```c
if( (curmat != MAT_WATER) ? S_SPLASH2 : (S_SPLASH1 >= 0) )
```

so I think this macro needs to be fixed by adding parentheses around the arguments:

```c
if((mw) >= 0)
```

voidanix: Awesome catch! Looks a bit ugly now though 👀 

Not even worth converting into a function because this macro is used twice in the codebase...

voidanix:
FYI the inrange function now uses `constexpr` (it will be reused in a later refactor) and the custom NULL macro has been removed

src/game/client.cpp
```diff
@@ -477,11 +477,15 @@ namespace client
...
         }
-        game::player1.loadweap.shrink(0);
-        loopv(items) if(game::player1.loadweap.find(items[i]) < 0)
-        {
-            game::player1.loadweap.add(items[i]);
-            if(game::player1.loadweap.length() >= W_LOADOUT) break;
+        game::player1.loadweap.clear();
+        for (const auto i : items) {
+            if (std::find(game::player1.loadweap.begin(),
+                game::player1.loadweap.end(), i) == game::player1.loadweap.end()) // NOT found
```
TheAssassin, line 483: I think the comment is not needed, even less experienced C++ programmers should know this idiom.

Also I'm pretty much the opposite of a fan of inline comments. They just increase the line length for no apparent reason and are hard to spot/find/read as well.

src/game/client.cpp
TheAssassin: It saves you a little bit of code, and also helps make clear what you're doing here. I guess you want to use `std::none_of`, though.

src/game/client.cpp
```diff
@@ -477,11 +477,15 @@ namespace client
...
+            {
+                game::player1.loadweap.emplace_back(i);
+                if (game::player1.loadweap.size() >= W_LOADOUT)
```
TheAssassin, line 486: Please wrap all conditional code (e.g., branches of an `if`) in braces to create a proper scope. It is a common issue in refactorings to add lines to existing code without realizing that the scope had ended at the semicolon. This can be avoided by always using a scope.

src/game/game.h
```diff
@@ -608,8 +606,9 @@ struct clientstate
 
     void addlastweap(int weap)
     {
-        lastweap.add(weap);
-        if(lastweap.length() >= W_ALL) lastweap.remove(0);
+        lastweap.emplace_back(weap);
+        if (lastweap.size() >= W_ALL)
```
TheAssassin, line 610: Same as above, please use a scope.

src/game/game.h
```diff
@@ -659,7 +658,8 @@ struct clientstate
             weapwait[i] = weaptime[i] = weapload[i] = weapshot[i] = prevtime[0] = 0;
             if(full) ammo[i] = entid[i] = -1;
         }
-        if(full) lastweap.shrink(0);
+        if (full)
```
TheAssassin, line 661: See above. Please revisit all your changes.

src/game/physics.cpp
```diff
@@ -1050,7 +1050,7 @@ namespace physics
             { \
                 int col = (int(mcol[2]*mq) + (int(mcol[1]*mq) << 8) + (int(mcol[0]*mq) << 16)); \
                 regularshape(mp, mt, col, 21, 20, mz, mo, ms, 1, 10, 0, 20); \
-                if(mw >= 0) playsound(mw, mo, pl); \
+                if((mw) >= 0) playsound(mw, mo, pl); \
```
TheAssassin, line 1053: What warning does this fix?

src/game/server.cpp
```diff
@@ -3395,7 +3395,8 @@ namespace server
                 }
                 prev.clear();
             }
-            if(!buf.empty()) setmods(sv_previousmaps, buf.c_str());
+            if(!buf.empty())
```
TheAssassin, line 3398: While you're at it, please add spaces in between statements such as `if` and `for` and the corresponding conditions.

We really need `clang-format` to auto check changes.

TheAssassin (changes requested):
I found a few more things.

Please also clean up the branch history when you're done.

src/game/server.cpp
robalni: The current style is to not have a space there. I'm not a big fan of mixing styles because the point of a style is the consistency that it gives.

src/game/client.cpp
voidanix: Would this work?:
```c++
if (std::none_of(game::player1.loadweap.begin(),
                game::player1.loadweap.end(), [=](auto d) { return d == i; }))
```
I hope there isn't too much overhead if done like this.

voidanix: > This can be avoided by always using a scope.

That is silly and adds unnecessary LoC; even the Kernel's coding style agrees:
> Do not unnecessarily use braces where a single statement will do.

src/game/physics.cpp
voidanix: Have you checked the commit? See https://github.com/blue-nebula/base/pull/248/commits/d0818f3992e248851eebf22a6deae86360b73642 and https://github.com/blue-nebula/base/pull/248#discussion_r905838467

src/game/server.cpp
voidanix: > The current style is to not have a space there. I'm not a big fan of mixing styles because the point of a style is the consistency that it gives.

That is the codebase's old style: we aren't fans of mixing stuff either but every new piece of code should be adopting the new style.

src/game/physics.cpp
TheAssassin: In that case it might just be removed entirely. That tiny bit of code duplication is acceptable.

src/game/client.cpp
TheAssassin: Just because the Linux kernel uses it doesn't mean it's beginner friendly. Quite the opposite, really. I've seen this kind of problem too often, so I always use a scope now. It's a pre-emptive measure used to prevent branching problems.

"Saving lines of code" is an argument which I often heard before, but it is not a good one. It might have been relevant, back in the days when there were just 12 lines on one screen. Compacting the code too much like this does not aid readability, quite the opposite, and is often used to compact and therefore kind of "hide" complexity. Look at this line, it's probably too long already.

Python for instance, while not everyone likes the lack of braces, does this very well in general. Whenever the code branches, moving the conditional code on a new line (and therefore adding some indentation as well) is very strongly recommended. You _could_ use the same line, but this is ugly and cumbersome. See https://docs.python-guide.org/writing/style/#one-statement-per-line.

Regarding braces, many popular code styles demand them, there are just few exceptions, mostly for old projects who had established such a style long ago. Not using braces "saves" 1-2 lines depending on the code style. I personally prefer to move the opening one on the same line as the conditional/loop statement. As stated in https://github.com/blue-nebula/base/pull/248#discussion_r917471888, this ensures that on future refactorings, new code won't accidentally be added in the hope that it will be conditional as well. Furthermore, this also aids the readability of the code. You can quickly see where branching happens. New variables have a clear scope. You can't be surprised when commenting out lines. You do not have to be careful about the semicolon being the end of the conditional code block.

This is not an obfuscated code competition. The code should be easy to read. Branches should be easily visible. Code should be written with the need for future changes in mind.

src/shared/tools.h
```diff
@@ -5,11 +5,7 @@
 #include <utility>
 #include <vector>
 #include <algorithm>
-
-#ifdef NULL
-#undef NULL
-#endif
-#define NULL 0
```
robalni, line 11: What is the reason to replace a well defined `NULL` value with an implementation defined value? As we can read on https://en.cppreference.com/w/cpp/types/NULL the value of `NULL` can be `0` or `nullptr`. These two values have different sizes on 64 bit computers and not knowing which one it is can cause problems in some situations, like when using varargs where you specify which type you want to read but `NULL` makes sure that you don't know which type you give to the function.

I think a better change (if we want to change it) might be to define it as `nullptr` or `(void *)0` or `(size_t)0` to make sure that it has the same size as a pointer because `NULL` is often used there you would use a pointer. Another solution could be to ban `NULL` (but then we would have to change in a lot of places) by defining it to a value that will generate a compilation error when used. That could be done if we want to avoid confusion that people think they are using the standard library `NULL`  when they are really using our `NULL`.

voidanix: First of all, this partly solves the `-Wzero-as-null-pointer-constant` warning which is (yet?) not enabled in BN.

> NULL is often used there you would use a pointer.

Almost always, actually: indeed using `nullptr` would totally eliminate the issue of weird conversions.

> like when using varargs where you specify which type you want to read but NULL makes sure that you don't know which type you give to the function.

I fail to see how different sizes actually makes things act weirdly, but using `nullptr` (which _must_ be passable through varargs) would be nicer too.

> Another solution could be to ban NULL

I encourage everybody to use `nullptr`, but some files are still using pure C here: once a conversion of sorts is completed, then sure.


FWIW, while my `NULL` expands to `__null`, I know that MS's (and MINGW's, according to libcxx) expands to `0`: see this commit as a stop gap, because I would much rather see some platforms spit out warnings about NULL than giving an error.

robalni: > I fail to see how different sizes actually makes things act weirdly

It's something I read somewhere but I'm not able to reproduce it so I don't know if it's actually a problem.

> but some files are still using pure C

If we make a new constant like `#define null ((void *)0)` that can be used in C too.

> my NULL expands to __null

And my `NULL` expands to `((void *)0)` on Debian/amd64/gcc :)

> I would much rather see some platforms spit out warnings about NULL than giving an error

Yes, it's probably a bad idea to hard ban `NULL` because there can be code (like in the standard library itself) that uses `NULL` that we don't easily control and then that code would break. For the same reason it might actually be a good idea to leave `NULL` defined as in the standard library because maybe there is some code in there that has expectations on the definition of `NULL`. I don't know.

voidanix:
Ping, any updates on this?

PR #255, opened 2023-04-03
bots no longer fire through walls

bmorel:


TheAssassin:
Seems to work, quite some improvement. More testing required, though.

bmorel:
it's a very old thing, really, I don't know why I never pushed this. Looks like I have bunch of refactoring branches on disk, not sure if worth the time to get those, I don't remember having more fixes/features

TheAssassin (approved):
Seems to work. I'd like to hear @voidanix's thoughts, though.

voidanix (approved):
Code style is a little broken with that 193 columns line, but otherwise LGTM and seems to supersede #185 as the bots "shake around" less when no target is in sight.

PR #261, opened 2023-04-20
cmake: Require OpenGL for clients only

voidanix:
When building the server only, CMake would yell that OpenGL is required
even if totally unnecessary: move the OpenGL requirement inside
BUILD_CLIENT where it should belong.


TheAssassin (approved):

TheAssassin:
I'll update #257 once this is merged, the build stage there installs OpenGL for no good reason.

TheAssassin:
Thanks!

PR #262, opened 2023-04-20
Support generation of NSIS installer with CPack

TheAssassin:
Will be used in the release repository to generate proper installers.


voidanix (changes requested):
Please update the ICO file path to what #263 defined, otherwise looks good.

voidanix:
While you are at it, can you also:

- Add `CPACK_PACKAGE_VENDOR`
- Replace `CPACK_PACKAGE_HOMEPAGE` with `CPACK_PACKAGE_HOMEPAGE_URL`

The possibility of using `CPACK_PACKAGE_ICON` (instead of `CPACK_NSIS_MUI_ICON`) should also be explored, especially if it could be used by both NSIS and future installers for other platforms.

TheAssassin:
Will do. Investigating the addition of `DragNDrop` support for macOS `.dmg` bundles, which makes use of some of these, too.

TheAssassin:
For the record, I already evaluated `CPACK_PACKAGE_ICON` extensively, it doesn't replace `CPACK_NSIS_MUI_ICON` and breaks other stuff. I'd stay away from it.

voidanix (approved):

PR #263, opened 2023-04-21
Rebranding, part deux

voidanix:
- cmake: nix: Update icon packaging
- cmake: nix: Add 256 and 512 icon variants
- install: win: Remove remnants of NSIS installer
- install: win: Move Windows-specific files from src
- scripts: Add icon generation script
- install: nix: Update icons
- install: win: Use BN ICO
- data: Update textures submodule


voidanix:
Hmm, seems to conflict with https://github.com/blue-nebula/base/pull/262: marking as WIP.

TheAssassin:
Please feel free to ignore #262, we'll fix that one after the icons have been fixed.

src/scripts/update-icons
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
```
TheAssassin, line 1: This file contains bashisms, so better request bash explicitly.

```suggestion
#! /bin/bash
```

src/scripts/update-icons
```diff
@@ -0,0 +1,14 @@
...
+#
+# This script updates the application icons define in install/nix/blue_nebula_x**.png from the original-sized icon.png using ImageMagick
+# The sizes we want that are used in the for loop are also defined in install_nix.cmake
+
+for i in 16 32 48 64 128 256 512; do
+    echo "Scaling icon to ${i}x${i}..."
+    convert -resize ${i}x${i} $1 ../install/nix/blue-nebula_x${i}.png
```
TheAssassin, line 8:
```suggestion
    convert -resize "${i}x${i}" "$1" "../install/nix/blue-nebula_x${i}.png"
```

    
Highly recommend using shellcheck to validate escaping.

src/scripts/update-icons
```diff
@@ -0,0 +1,14 @@
...
+    if [ "$i" -eq 512 ]; then
+        cp ../install/nix/blue-nebula_x${i}.png ../../data/textures/icon.png
```
TheAssassin, line 10:
```suggestion
        cp "../install/nix/blue-nebula_x${i}.png" ../../data/textures/icon.png
```

src/scripts/update-icons
TheAssassin, line 1: Consider adding a `.sh` suffix.

src/install/win/blue-nebula.rc
TheAssassin, line 1: I think the current purpose of this `src/install` directory is to house files needed to generate the installer. The `.rc` file is required during compilation to include icons into the file. However, the `.ico` is again in `src/install/win`, so there is no clear pattern. We might want to reorganize files, and possibly rename `install` to `platform` or `native` or something like that.

src/cmake/install_nix.cmake
```diff
@@ -13,14 +13,14 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
...
             COPY ${CMAKE_CURRENT_SOURCE_DIR}/install/nix/blue-nebula_x${res}.png
-            DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/icons/hicolor/${res}/apps/
+            DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/icons/hicolor/${res}x${res}/apps/
         )
         file(RENAME
-            ${CMAKE_CURRENT_BINARY_DIR}/icons/hicolor/${res}/apps/blue-nebula_x${res}.png
-            ${CMAKE_CURRENT_BINARY_DIR}/icons/hicolor/${res}/apps/blue-nebula.png
+            ${CMAKE_CURRENT_BINARY_DIR}/icons/hicolor/${res}x${res}/apps/blue-nebula_x${res}.png
+            ${CMAKE_CURRENT_BINARY_DIR}/icons/hicolor/${res}x${res}/apps/blue-nebula.png
             )
```
TheAssassin, line 24: While we're at it, mind to fix the code style?

```suggestion
        )
```

TheAssassin (changes requested):

src/scripts/update-icons
voidanix: Where does it do so? It runs perfectly fine with dash, plus both shellcheck and checkbashisms do not say anything.

src/scripts/update-icons
voidanix: I suppose this must be done for almost every file in scripts though?

src/scripts/update-icons
TheAssassin: Seems like I was wrong, I updated shellcheck. In any case, I think using bash makes sense, as it allows you to `set -euo pipefail`.

src/scripts/update-icons
TheAssassin: Well, in the long term, I think that would be a good idea. Feel free to ignore my suggestion for now.

TheAssassin (approved):

TheAssassin:
Thanks!

PR #272, opened 2025-06-09
Add serverpublicip variable and auto-publish Docker images

TheAssassin:


src/game/server.cpp
```diff
@@ -7321,6 +7321,7 @@ namespace server
         {
             case 'P': setsvar("adminpass", &arg[3]); return true;
             case 'k': setsvar("serverpass", &arg[3]); return true;
+            case 'j': setsvar("serverpublicip", &arg[3]); return true;
```
voidanix: This is pretty ugly...

.gitignore
```diff
@@ -32,6 +32,7 @@ out
 *.core
 *.diff
 *.patch
+!docker/serverip.patch
```
voidanix: Already removed.

voidanix (changes requested):
Commits would need to be cleaned up if possible: only one nit, otherwise the rest looks good

PR #274, opened 2025-12-05
Update appstream, .desktop file and add appstream install step in cmakelists

user1-github:
As promised, this is the main PR with the updates / improvements I was talking about.
Note: before merging it, [this PR](https://github.com/blue-nebula/misc/pull/1) in the misc repo should be merged, so that I can change to the proper screenshot urls in the appstream file.

Before releasing a new version, the new version number should be added to the releases section of the appstream file.

Btw, I've also added brand colors to the appstream file. It's not a hard requirement, but an optional aesthetic touch to make an app look nice on various Linux store frontends. Here is how they look like:

Blue_Nebula_brand_colors

@voidanix Regarding the Flatpak, based on my changes above, I've already built a [working test build](https://github.com/flathub/net.redeclipse.RedEclipse-Legacy/pull/35#issuecomment-3617366516) of Blue Nebula Flatpak and it works well! (ignore the Red Eclipse Legacy id, as I cannot change it in PRs of existing apps).
Also, [here is how the Flatpak manifest currently looks like](https://github.com/user1-github/net.redeclipse.RedEclipse-Legacy/blob/blue-nebula-test/net.redeclipse.RedEclipse-Legacy.yml). It turned out to be much simpler than I thought it would be thanks to the well made cmake system install steps.

So now I guess the only thing left is making the first official release after which we can publish it on Flathub :)

Though I still wonder if we should rebase the Red Eclipse Legacy Flatpak to Blue Nebula Flatpak.
Because at this point Blue Nebula is basically an improved RE 1.6.0, so I don't really see how would it disappoint some users. Afaik, the eol rebase doesn't force users to switch to the new Flatpak (it first asks them) and if they agree, then there's also some mechanism to migrate user data from the old Flatpak to the new one.
Or is there something 1.6.0 used to have that Blue Nebula has removed?


src/install/nix/blue-nebula.metainfo.xml.am
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>@APPNAME@.desktop</id>
+  <launchable type="desktop-id">@APPNAME@.desktop</launchable>
+  <metadata_license>CC-BY-SA-3.0+</metadata_license>
+  <project_license>Zlib and CC-BY-SA-3.0+</project_license>
+  <name>Blue Nebula</name>
+  <developer_name>Blue Nebula Team</developer_name>
```
voidanix: Use the non-deprecated `developer` tag instead:

```suggestion
<developer id="org.blue-nebula">
  <name>Blue Nebula Team</name>
</developer>
```

src/install/nix/blue-nebula.metainfo.xml.am
```diff
@@ -0,0 +1,52 @@
...
+      <li>Features parkour, impulse boosts, dashing, and other tricks</li>
+      <li>Favourite gamemodes with an array of mutators and variables</li>
+      <li>Built-in editor lets you create your own maps cooperatively online</li>
+    </ul>
+  </description>
+  <url type="homepage">https://blue-nebula.org</url>
+  <url type="bugtracker">https://github.com/blue-nebula/base/issues</url>
+  <screenshots>
+    <screenshot type="default">
+      <image>https://raw.githubusercontent.com/user1-github/misc/master/screenshots/053.jpg</image>
```
voidanix: These should already be pointing to the misc repo

voidanix (changes requested):

user1-github:
Done, I have 3 questions (slightly unrelated to this PR):

- I need to choose an app id before I submit the Flatpak. Is the following id ok for you? `org.blue_nebula.BlueNebula`

- In the Flatpak manifest I have to set `-DCMAKE_POLICY_VERSION_MINIMUM=3.5`, because Flatpak builder doesn't accept anything older. Should we update the minimum cmake version in cmakelists, or is there a good reason to stay with the current version?

- I also think it's a good idea to update enet to the latest version for potential bugfixes. Though I can't deal with git submodules atm, so I won't be able to update it myself.

user1-github:
>FWIW I believe you should also update the developer id with the underscore to make things consistent.

I also thought about it, but from what I've seen in other apps, I think the developer id should be left as is. For example, [Dosbox X Flatpak](https://github.com/flathub/com.dosbox_x.DOSBox-X/blob/master/com.dosbox_x.DOSBox-X.metainfo.xml#L10) has an underscore in its app id, but not in the developer id. I just don't know if changing to an underscore in developer id will cause issues.

voidanix:
> For example, Dosbox X Flatpak has an underscore in its app id, but not in the developer id.

Good to know. Fine by me then :)

PR #290, opened 2025-12-25
Simplify draw functions (courtesy of @robalni)

TheAssassin:
Thanks @robalni for PR #218! Unfortunately, the fork was deleted, so we could not merge it without creating a new PR.


TheAssassin:
I share your concern, but it's more readable than the original anyway.

